PR #42, opened 2014-12-22
Added option to ignore expirations in tokens

ben-bradley:
I added code and tests to allow for the ignoring of expired tokens.


jfromaniello:
Thanks this will be released today as v4

ben-bradley:
Sweet, thanks! =)

PR #50, opened 2015-01-23
Fixed exp value when noTimestamp is required.

santiagoaguiar:


jfromaniello:
Thank you this will be release on v4 today.

PR #55, opened 2015-02-14
Changes aud/iss error to use actual expected value

abalmos:
The message part of the JsonWebTokenError generated for `aud` and `iss` mismatch
currently use the [PAYLOAD AUDIENCE] and [PAYLOAD ISSUER] as the expected value.
This leads to confusion.  For example, say a JWT `aud` is set to
`https://localhost` and the expected value is `https://localhost:8443`. The
resulting error message is:

`jwt audience invalid. expected: https://localhost:8443`

Which of courses tells the user that the audience is incorrect, yet the expected
value is the value sent.

This commit changes the error message to use the actual expected values,
[OPTIONS AUDIENCE] and [OPTIONS ISSUER].

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,8 +162,8 @@ Error object:
   * 'jwt malformed'
   * 'jwt signature is required'
   * 'invalid signature'
-  * 'jwt audience invalid. expected: [PAYLOAD AUDIENCE]'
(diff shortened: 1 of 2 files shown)


dschenkelman:
This makes absolute sense, thank you!

PR #56, opened 2015-02-20
Assumes jwtString is a JWT

abalmos:
JWS checks for `type: "JWT"` in the JWS header to determine if it should JSON.parse the payload or not. The standard makes this key optional. jsonwebtoken already makes a strong assumption that the JWS is a JWT so we should pass that assumption on the JWS.

### DIFF
--- a/index.js
+++ b/index.js
@@ -100,7 +100,7 @@ module.exports.verify = function(jwtString, secretOrPublicKey, options, callback
   var payload;
 
   try {
-   payload = this.decode(jwtString);
+   payload = this.decode(jwtString, {json: true});
   } catch(err) {
     return done(err);
   }


jfromaniello:
I think we are fine with the current behavior, if `typ: 'JWT'` decode will return string.

abalmos:
@jfromaniello node-jsonwebtoken is broken without some sort of change.

If a JWT that is being validated does not have a typ: 'JWT' key in its header (allowed by JWT spec) then jws returns a string rather than a parsed object. node-jsonwebtoken will only validate the claim set if the decoded value is an object. Therefore, valid JWTs with a JSON claim set but no typ: 'JWT' in the header are not correctly validated but are valid JWTs. See section 7.2, step 10 of draft-ietf-oauth-json-web-token

jfromaniello:
Can you send a test?

abalmos:
@jfromaniello Sorry, forgot to mention I made #63 which handles the issue without breaking any existing tests. It also has a test for the no `typ: 'JWT'` case.

PR #57, opened 2015-02-27
Update README to document the headers option.

drichard:
Document the headers option in the jwt.sign method.


jfromaniello:
thanks!

PR #59, opened 2015-03-02
Update jws to 2.0.0 [UTF-8 encoding support]

RoberMac:
In ```node-jws``` [v2.0.0](https://github.com/brianloveswords/node-jws/blob/master/CHANGELOG.md#200---2015-01-30) , default payload encoding changed from ```binary``` to ```utf8```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jsonwebtoken",
-  "version": "3.2.2",
+  "version": "3.2.3",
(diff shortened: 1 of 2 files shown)


chrisprice:
I've just stumbled across this one as well. :+1:

jfromaniello:
Thanks! I did the same on #62 but I allowed to set the encoding format. this will be a major release 4.

Since it can break existing code.

Thanks again

PR #63, opened 2015-03-06
First assume JWT claim set is an object

abalmos:
draft-ietf-oauth-json-web-token does not require JWT objects to have typ: "JWT"
in their header. Yet it expects that the JWT's claim set be JSON for validation.
See step 10 of draft-ietf-oauth-json-web-token section 7.2 "Validating a JWT".

Prior to this commit JWTs without typ: "JWT" in header are validated without
ever checking the claim set.


jfromaniello:
Thanks ! it makes sense. I published it as `v4.1.0`.

Issue #64, opened 2015-03-10
Options ignored when payload isn't an object


aiguofer:
According to the docs, the payload can be a string, buffer, or object. However, the `sign` function blindly tries to add the `exp`, `iss`, `aud`, and `sub` properties to the payload, so when it's a primitive those properties get lost.

I'm submitting a pull request for a possible fix, although there might be a need for better type checking/handling.

jfromaniello:
what would you expect in this case?

aiguofer:
I would expect that setting options would have the same effect for any *allowed* payload and throw some sort of error or warning otherwise. 

The JWT draft isn't very clear about this, although from reading it I get the impression that the payload should only be allowed to be an object (or a Claims set as they call it). If that's the case, then it should fail when you try to pass a string as the payload. Otherwise, maybe the string can be put into a "content" claim inside the claims set (like I did in the pull request).

What do you think? Should this be brought to the authors of the standard to see what they think?

jfromaniello:
I think throwing is a the way to go. If you are `jwt.sign('foo', {audience: 'bar'}` it should throw because the payload can't have audience

aiguofer:
Yeah... I think based on the current definition that would make the most sense. 

However, it feels like options should be separate from content within the payload (or just separate entities altogether), but that seems like conversation for the authors of the standard and not any specific implementation.

davidnknight:
I just got caught out by this myself. I am fetching a user object from the database and passing that user object back as the payload, to which options were completely ignored and it wasn't clear why until I came across this issue, and realised that it was trying to append options as attributes to the payload object.

To throw some sort of exception, or even just a console logged warning that your options can't be appended would be useful rather than noticing token expiry is being ignored, to then have to spend time debugging and working out why.

MariusRumpf:
I wanted to bring back attention back to this issue. This should be rather easy to fix, by throwing an exception for strings or checking the type of argument passed and converting it to object I would think. 
This issue seems to gain lot of popularity lately and is somewhat a little security problem if one assumes the tokens are expiring but they aren't. Same for other possible options. 
 Is their any plan to address this issue?

jfromaniello:
@MariusRumpf yes, I will thrown an error if the user provides `expiresIn`, `issuer`, `audience` etc and the payload is not an object in the next major version

jfromaniello:
fixed in v6. 
thanks

PR #65, opened 2015-03-10
Fix payload properties when not an object

aiguofer:
Created this to go along with issue #64. I wasn't sure what to do with 
```
if (typeof pl.content === 'object') {
  header.typ = 'JWT';
}
```

Maybe it should just be set to JWT regardless? The spec seems vague about it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,9 @@ module.exports.sign = function(payload, secretOrPrivateKey, options) {
 
   var header = ((typeof options.headers === 'object') && options.headers) || {};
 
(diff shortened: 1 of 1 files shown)


aiguofer:
Well nvm, this fix assumes that you'd only want the payload content after it's been verified and not any of the options (which causes many of those tests to fail). I'm not sure what the preferred method for a fix would be, but here's an idea :)

aiguofer:
As per the discussion in the issue, it's better to throw an error when trying to set options on a string payload.

PR #69, opened 2015-03-16
Verification with an asymmetric key of a token signed with a symmetric key

jfromaniello:
There is a vulnerability in this module when the verification part is expecting a token digitally signed with an asymetric key (RS/ES family) of algorithms but instead the attacker send a token digitally signed with a symmetric algorithm (HS* family).

The issue is because this library has the very same signature to verify both type of tokens (parameter: `secretOrPublicKey`).

This change adds a new parameter to the verify called `algorithms`. This can be used to specify a list of supported algorithms, but the default value depends on the secret used: if the secretOrPublicKey contains the string `BEGIN CERTIFICATE` the default is:
```
[ 'RS256','RS384','RS512','ES256','ES384','ES512' ]
``` 
otherwise is 
```
[ 'HS256','HS384','HS512' ]
````

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -15,6 +15,7 @@
     "it": true,
     "require": true,
     "atob": false,
-    "escape": true
+    "escape": true,
+    "before": true
   }
 }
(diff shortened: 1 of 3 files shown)


jfromaniello:
published as `v4.2.0`.

pose:
@jfromaniello Thanks for fixing this :)

Issue #70, opened 2015-03-17
RS256 problem 4.2.x


graovic:
Hi,

this package have a problem with decoding RS256 signed messages

var jwt = require('jsonwebtoken');
var fs = require('fs');
var cert_pub = fs.readFileSync('app.rsa.pub');
var cert_priv = fs.readFileSync('app.rsa');

var token = jwt.sign({ foo: 'bar' }, cert_priv, { algorithm: 'RS256'});
jwt.verify(token, cert_pub, function(err, decoded) {
    console.log("Decoded: " + JSON.stringify(decoded));
    console.log("Error: " + err);
});

This code woks fine in 4.1.0, but on 4.2.0 version problem is with verifying token.

jfromaniello:
What's the error you get? It seems to be working fine for me. Check the test I added on the branch issue_70.

graovic:
~/ cat package.json| grep json
    "jsonwebtoken": "4.1.0",
~/ node testtoken.js  
Decoded: {"foo":"bar","iat":1426605073}
Error: null

 ~/ cat package.json| grep json
    "jsonwebtoken": "^4.2.0",
~/ node testtoken.js
Decoded: undefined
Error: JsonWebTokenError: invalid signature

jfromaniello:
can you give me the first line of your app.rsa.pub file? thanks

graovic:
-----BEGIN PUBLIC KEY-----

jfromaniello:
got it! Thanks, it should be fixed on 4.2.1

jedwards1211:
uhhh...my `id_rsa.pub` looks different:

```
ssh-rsa <KEY> <TOKEN>
```

Should this format be supported by jsonwebtoken?

PR #71, opened 2015-03-26
Fix verify for RSAPublicKey formated keys

awlayton:
Ran into a problem like in #70, but with a public key beginning with:
```
-----BEGIN RSA PUBLIC KEY-----
```

This fixes my issue.


jfromaniello:
Thank you! published as 4.2.2

PR #72, opened 2015-03-30
Empty secret or public key

JackuB:
When secret or key fails to load or is not provided, `verify` can throw error even with callback.

Added check for `null` and `undefined`

### DIFF
--- a/index.js
+++ b/index.js
@@ -107,6 +107,9 @@ module.exports.verify = function(jwtString, secretOrPublicKey, options, callback
     return done(new JsonWebTokenError('jwt signature is required'));
   }
 
+  if (typeof secretOrPublicKey === "undefined" || secretOrPublicKey === null) // secretOrPublicKey can be empty string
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -107,6 +107,9 @@ module.exports.verify = function(jwtString, secretOrPublicKey, options, callback
     return done(new JsonWebTokenError('jwt signature is required'));
   }
 
+  if (typeof secretOrPublicKey === "undefined" || secretOrPublicKey === null) // secretOrPublicKey can be empty string
```
jtwebman: Can't we change this to just:
```javascript
if (!secretOrPublicKey) {
}
```

JackuB: @jtwebman I got impression, that key is allowed to have value of `''` (empty string).

Seems safer to check even for empty string, updated code.

jfromaniello:
Thank you very much, published as `v5.0.4`.

PR #75, opened 2015-04-07
Only set iat if there is none

dschenkelman:
Makes sense?


pose:
@dschenkelman Looks good to me! :+1:

PR #78, opened 2015-04-10
Update to jws@^3.0.0

pose:
As `jws@3.0.0` changed the verify method signature to be `jws.verify(signature, algorithm, secretOrKey)`, the token header must be decoded first in order to make sure that the `alg` field matches one of the allowed `options.algorithms`. After that, the now validated `header.alg` is passed to `jws.verify`

As the order of steps has changed, the error that was thrown when the JWT was invalid is no longer the `jws` one:

```
{ [Error: Invalid token: no header in signature 'a.b.c'] code: 'MISSING_HEADER', signature: 'a.b.c' }
```

That old error (removed from jws) has been replaced by a `JsonWebTokenError` with message `invalid token`. That's why this change will bump be a major.

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,11 +117,13 @@ module.exports.verify = function(jwtString, secretOrPublicKey, options, callback
 
   }
 
-  if (!jws.isValid(jwtString)) {
+  var decodedToken = jws.decode(jwtString);
+
+  if (!decodedToken) {
     return done(new JsonWebTokenError('invalid token'));
   }
 
-  var header = jws.decode(jwtString).header;
+  var header = decodedToken.header;
 
   if (!~options.algorithms.indexOf(header.alg)) {
     return done(new JsonWebTokenError('invalid signature'));
--- a/index.js
+++ b/index.js
@@ -112,15 +112,25 @@ module.exports.verify = function(jwtString, secretOrPublicKey, options, callback
                          ~secretOrPublicKey.toString().indexOf('BEGIN PUBLIC KEY') ?
                           [ 'RS256','RS384','RS512','ES256','ES384','ES512' ] :
                          ~secretOrPublicKey.toString().indexOf('BEGIN RSA PUBLIC KEY') ?
-							[ 'RS256','RS384','RS512' ] :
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -112,15 +112,25 @@ module.exports.verify = function(jwtString, secretOrPublicKey, options, callback
...
+                          [ 'RS256','RS384','RS512' ] :
+                          [ 'HS256','HS384','HS512' ];
 
   }
 
+  if (!jws.isValid(jwtString)) {
+    return done(new JsonWebTokenError('invalid token'));
+  }
+
+  var header = jws.decode(jwtString).header;
```
dschenkelman: Does it make sense to check validity and decode? What if we decode and handle the failure?

dschenkelman:
@pose couple of questions:
* Do we need more tests for this change?
* Are there any other things we would like to get in for the next major that would also break the API?

index.js
pose: Yes, you are right. If I do `jws.decode` and check for null is the same.

pose:
> Do we need more tests for this change?
We have a couple of tests of invalid tokens: https://github.com/auth0/node-jsonwebtoken/blob/update-to-jws-3/test/jwt.rs.tests.js#L239-L278 Do you have any additional test in mind?

> Are there any other things we would like to get in for the next major that would also break the API?
No that I'm aware of. I've taken a look at the PRs and issues.

index.js
```diff
@@ -112,15 +112,27 @@ module.exports.verify = function(jwtString, secretOrPublicKey, options, callback
...
+  var decodedToken = jws.decode(jwtString);
+
+  if (!decodedToken) {
+    return done(new JsonWebTokenError('invalid token'));
+  }
+
+  var header = decodedToken.header;
+
+  if (!~options.algorithms.indexOf(header.alg)) {
+    return done(new JsonWebTokenError('invalid signature'));
```
dschenkelman: @jfromaniello @pose should this be `'invalid header'` or `'invalid 'alg' claim`?

PR #81, opened 2015-04-19
Fix link to auth0.com in README

jas:
The README currently links to https://github.com/auth0/node-jsonwebtoken/blob/master/auth0.com rather than https://auth0.com.


pose:
Good catch @jas, thanks!

Issue #14, opened 2014-07-15
An exception is thrown when payload is a string


asafyish:
When using jwt.verify or jwt.decode with a payload that was created from a string, there is an exception.

``` javascript
var jwt = require('jsonwebtoken');
var token = jwt.sign('hello', '123');
jwt.verify(token, '123', function(err, decoded) {
    // Will never run because of 'SyntaxError: Unexpected token'
    console.log(decoded);
});
```

According to the docs for sign() - "payload could be an literal, buffer or string".

PR #88, opened 2015-05-07
Avoid uncaught "SyntaxError: Unexpected token ͧ" error.

dbrugne:
When .verify() a corrupted JWS (e.g.: malicious user add extra characters in the middle of token) an error is thrown:

``
[ERROR] console - SyntaxError: Unexpected token ͧ
    at Object.parse (native)
    at Object.jwsDecode [as decode] (/www/socketio-jwt/node_modules/jsonwebtoken/node_modules/jws/lib/verify-stream.js:71:20)
    at Object.module.exports.verify (/www/socketio-jwt/node_modules/jsonwebtoken/index.js:120:26)`
 ``

Fix is similar to: https://github.com/auth0/node-jsonwebtoken/issues/14

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,7 +117,12 @@ module.exports.verify = function(jwtString, secretOrPublicKey, options, callback
 
   }
 
-  var decodedToken = jws.decode(jwtString);
+  var decodedToken;
+  try {
+    decodedToken = jws.decode(jwtString);
+  } catch(err) {
+    return done(new JsonWebTokenError('invalid token'));
+  }
 
   if (!decodedToken) {
     return done(new JsonWebTokenError('invalid token'));


forlooper:
I encountered this error as well. Right now I had to write my code like this:

```
try {
    jwt.verify(token, secret, function (err, decoded) {
    }
}
catch(e) {
    //error caught
}

```

The verify method could pass an err in the callback. Why not use that?

dbrugne:
@forlooper , I don't understand your comment.

*"The verify method could pass an err in the callback. Why not use that?"*

This is exactly what was implemented in this pull request. Thrown exception capturing and passing to callback.

jfromaniello:
Thank you very much, published as `5.0.2`.

PR #89, opened 2015-05-15
Add option to return header and payload when decoding.

jaredhanson:
This adds an option to return the header and payload when decoding a token.  The header includes claims such as `kid`, which is needed to support RS* signatures and key rotation, where keys are found in a JWKS.

This is needed for a forthcoming patch to express-jwt, in order to give the header claims to the `secretCallback`

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,12 +137,18 @@ jwt.verify(token, cert, { algorithms: ['RS256'] }, function (err, payload) {
 `options`:
 
 * `json`: force JSON.parse on the payload even if the header doesn't contain `"typ":"JWT"`.
+* `complete`: return an object with the decode payload and header.
 
 Example
 
 ```js
 // get the decoded payload ignoring signature, no secretOrPrivateKey needed
 var decoded = jwt.decode(token);
+
+// get the decoded payload and header
+var decoded = jwt.decode(token);
+console.log(decoded.header);
+console.log(decoded.payload)
 ```
 
 ## Errors & Codes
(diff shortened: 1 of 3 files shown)


woloski:
should we document this option on the readme?

jaredhanson:
Added documentation to README.

README.md
```diff
@@ -137,12 +137,18 @@ jwt.verify(token, cert, { algorithms: ['RS256'] }, function (err, payload) {
...
+* `complete`: return an object with the decode payload and header.
 
 Example
 
 ```js
 // get the decoded payload ignoring signature, no secretOrPrivateKey needed
 var decoded = jwt.decode(token);
+
+// get the decoded payload and header
+var decoded = jwt.decode(token);
```
dschenkelman: should this be passing `{complete: true}` in order to use header?

test/jwt.rs.tests.js
```diff
@@ -275,5 +275,15 @@ describe('RS256', function() {
       assert.deepEqual(payload, obj);
       done();
     });
+    it('should return the header and payload and signature if complete option is set', function(done) {
+      var obj     = { foo: 'bar' };
+      var token   = jwt.sign(obj, priv, { algorithm: 'RS256' });
+      var decoded = jwt.decode(token, { complete: true });
+      console.log(decoded);
```
dschenkelman: remove

README.md
jaredhanson: Yes. My mistake

jaredhanson:
Updated the PR to fix what you pointed out.  Thanks!

PR #102, opened 2015-06-26
Added support for sub, jti and nbf claims

TATDK:


omsmith:
I'm not sure I quite understand the use-case behind asserting a particular _jti_. Could you enlighten me @TATDK?

TATDK:
From [RFC 7519 - JSON Web Token (JWT)](https://tools.ietf.org/html/rfc7519#section-4.1.7)

> The "jti" (JWT ID) claim provides a unique identifier for the JWT.
> The identifier value MUST be assigned in a manner that ensures that
> there is a negligible probability that the same value will be
> accidentally assigned to a different data object; if the application
> uses multiple issuers, collisions MUST be prevented among values
> produced by different issuers as well.  The "jti" claim can be used
> to prevent the JWT from being replayed.  The "jti" value is a case-
> sensitive string.  Use of this claim is OPTIONAL.

I mainly implemented these to ensure compatibility when communicating with other libraries that supports these.

focusaurus:
+1

aulvi:
Any update on this pull? I'd love to see this cleaned up and merged.

ryanwalters:
+1

holm:
:+1:

jurriaan:
+1

cgalibern:
+1

ghemingway:
+1

jfromaniello:
I did some changes to the final implementation:

1. the option for the `sign` method is called `notBefore`. It supports the same format than `expiresIn`. You can pass a number of seconds or an string like `6m` (six minutes), `5h` (fiver hours), etc.
2. the `verify` method doesn't have a jti options. This is on purpose, application don't normally validate a JWT with a 1-to-1 matching of JTI. You might either validate that the JTI is on a whitelist or isn't on a blacklist.

Issue #87, opened 2015-05-06
Missing documentation for expiresInMinutes and expiresInSeconds


samholmes:
Documentation should explain what possible values are valid for these options. Should the expiresInMinutes be the number of minutes from unix time or number of minutes from token generation?

Zertz:
The name implies it is from token generation: [token] expires in x minutes

jfromaniello:
Yes, it is from token generation and I don't like these parameters.

knownasilya:
Please make only one, it's a big pain if exposing the config to users because it's confusing which one will have effect.. One option should be enough.

jfromaniello:
These parameters are deprecated and will be removed on next major version.

alexilyaev:
For those who stumble upon this, we should use `expiresIn` from this point on.

PR #109, opened 2015-07-15
deprecate expireInMinutes and expireInSeconds - in favor of expires

jfromaniello:
This change adds a deprecation warning on the confusing `expiresInMinutes` and `expiresInSeconds` options of the sign method.

The new option `expires` accept seconds or an string that represents an amount of time. Examples:

```javascript
jwt.sign({foo: 'bar'}, 'my-secret', { expiresIn: 24 * 60 * 60 });
jwt.sign({foo: 'bar'}, 'my-secret', { expiresIn: '2d' });
jwt.sign({foo: 'bar'}, 'my-secret', { expiresIn: '2 days' });
jwt.sign({foo: 'bar'}, 'my-secret', { expiresIn: '10 hours' });
jwt.sign({foo: 'bar'}, 'my-secret', { expiresIn: '10h' });
```

This should also solve the concern expressed in #87.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ encoded private key for RSA and ECDSA.
 `options`:
 
 * `algorithm` (default: `HS256`)
-* `expiresInMinutes` or `expiresInSeconds`
+* `expiresIn`: expressed in seconds or an string describing a time span [rauchg/ms](https://github.com/rauchg/ms.js). Eg: `60`, `"2 days"`, `"10h"`, `"7d"`
 * `audience`
 * `subject`
 * `issuer`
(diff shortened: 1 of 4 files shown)


dschenkelman:
I like it. Can we add tests for good errors in case the string is not valid?

jfromaniello:
@dschenkelman que tal esto?

cristiandouce:
:laughing:

rolodato:
Just a minor phrasing change: I'd use `expiresIn` since `expires` is similar to `exp`, which is an absolute datetime instead of a time interval. Maybe have both `expiresIn` and `expiresOn`?

dschenkelman:
I like `expiresIn`

jfromaniello:
@dschenkelman  Heads up, I just renamed to `expiresIn` and I think is ready to merge.

gdl276s:
Is there a reason we are dividing by 1000 in several places as apposed to multiplying by 1000 when setting the expiration?  It would seem we are losing a bit of precision here (even if it is a small amount)

jfromaniello:
This change has been merged and released as `v@5.4.0`.

lvarayut:
It's good to put this breaking change in `CHANGELOG.md`

PR #110, opened 2015-07-16
Adds .verify option maxAge

jden:
A feature proposal presented for your consideration, with proposed tests and implementation.

As a JWT consumer, I would like to verify that the claims I receive have been issued within a certain timespan no more than maxAge ago, so that I can limit the exposure window during which the JWT may have been exposed to another party.

Think of it as a verifier-side expiration policy. In an integration scenario, the JWTs I'm consuming do not have an expiration, but they do have an `iat` claim.

Usage example:
```js
jwt.verify(
  'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJmb28iOiJiYXIiLCJpYXQiOjE0MzcwMjI2MzR9.wNkbZPARjNc1fKEvRzZxocw5S8LZvQx3EjT1BswKSOA',
  'key',
  {algorithms: ['HS256'], maxAge: '10s'}
)
// throws TokenExpiredError: maxAge exceded
```
Implementation note: to align with #109, `maxAge` does not have a unit in the name and instead is implemented using [ms](http://npm.im/ms). As of the time this PR was opened, #109 had not been merged. I will rebase my branch on that once it is merged.

### DIFF
--- a/index.js
+++ b/index.js
@@ -201,7 +201,7 @@ JWT.verify = function(jwtString, secretOrPublicKey, options, callback) {
       return done(new JsonWebTokenError('iat required when maxAge is specified'));
     }
     if (Date.now() - (payload.iat * 1000) > maxAge) {
-      return done(new TokenExpiredError('maxAge exceded', new Date(payload.iat * 1000 + maxAge)));
+      return done(new TokenExpiredError('maxAge exceeded', new Date(payload.iat * 1000 + maxAge)));
     }
   }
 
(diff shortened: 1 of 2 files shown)


jfromaniello:
I think is a great idea, perfectly implemented. :+1:

jfromaniello:
@jden can we support passing `maxAge` as an int in addition to string? as in #109

twistedstream:
@jden Great PR! Could you also update the README.md so the documentation shows this new option? You could also explain the use case vs. just using normal token expiration.

jfromaniello:
Published as `v5.2.0`. Thank you very much and sorry for the delay

PR #114, opened 2015-07-23
Add refresh token functionality

jmlaya:
Simple functionality to implement refresh token.

### DIFF
--- a/index.js
+++ b/index.js
@@ -197,3 +197,39 @@ JWT.verify = function(jwtString, secretOrPublicKey, options, callback) {
 
   return done(null, payload);
 };
+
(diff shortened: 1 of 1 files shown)


dschenkelman:
Hey @jmlaya! Thanks for the contribution.

I think this functionality would add something to the interface of this module that is not required. I believe a separate module that is composed with this one and performs this operation would be a better alternative.

Thanks!

rolodato:
I agree with @dschenkelman. I think this adds unnecessary complexity to the library for a feature that's simple to implement for most cases, which is just signing the same payload with a different expiration date.

PR #116, opened 2015-08-08
Fix options.headers for the sign method

simov:
First of all `options.header` was never hit, if you follow the documentation and set your headers via `options.headers`:

```js
if (options.header) {
    Object.keys(options.header).forEach(function (k) {
      header[k] = options.header[k];
    });
  }
```

That would be easy to spot if you had a test coverage tool enabled in your build.

Next - even if you add **s** at the end of `options.header` you still can't override the `alg` header, because having this code:

```js
var header = ((typeof options.headers === 'object') && options.headers) || {};
header.alg = options.algorithm || 'HS256';
```

Actually modifies the `options.headers` before you have the chance to read its value. That's because `headers` is an object (pointer), not a value type.

So the test is pretty simple:

```js
jwt.sign({}, '', {headers:{alg:'RS256'}})
```

That's never going to work!

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,7 +38,7 @@ JWT.decode = function (jwt, options) {
 JWT.sign = function(payload, secretOrPrivateKey, options) {
   options = options || {};
 
-  var header = ((typeof options.headers === 'object') && options.headers) || {};
+  var header = {};
 
   if (typeof payload === 'object') {
     header.typ = 'JWT';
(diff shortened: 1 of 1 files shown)


jfromaniello:
Thanks man, I just added two tests cases and pushed as `v5.0.5`.

Sorry for the delay

omsmith:
@jfromaniello Just so you're aware, this was a breaking change (change the name of the option), which you just released as a patch version.

omsmith:
@jfromaniello Actually, apologies. While the `headers` option is what was documented, we were using `header` because that's what actually worked (were only using it in tests, so wasn't too worried about it).

This is obviously correct as a patch, because it's updated to work as documented.

Thanks very much

simov:
Yep, basically what [jws](https://github.com/brianloveswords/node-jws) offers as interface is just fine:

```js
jws.sign({
  header: { alg: 'HS256' },
  payload: 'h. jon benjamin',
  secret: 'has a van',
});
```

Not to mention that's what the spec outlines.

I'm not exactly aware of what `jsonwebtoken` offers, but IMO the input parameters for the `sign` method are just unnecessary abstraction. Also it's good to have that interface working in case you are building on top of it.

Anyway, it works as documented now :+1:

PR #117, opened 2015-08-13
note that this library is secure

coolaj86:
I think it's important to note that this library is secured from known attacks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,12 @@ ES384 | ECDSA using P-384 curve and SHA-384 hash algorithm
 ES512 | ECDSA using P-521 curve and SHA-512 hash algorithm
 none | No digital signature or MAC value included
 
+## Security
+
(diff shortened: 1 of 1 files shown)


Apercu:
Indeed it may be interesting to note that fact, but it don't really agree with the *happy* term you've used. People using a crypto library expect it to be secure, especially when the issue has been announced long time ago publicly, and this should be the standard. Your sentence can imply that people should be good with what they have, even though it's a critical vulnerability.

rolodato:
The status of each JWT library respect to this vulnerability is already documented at http://jwt.io. I don't see the point of explicitly saying that a library is secure, because that is the default expectation for everyone. Also, this was already mentioned in the changelog here: https://github.com/auth0/node-jsonwebtoken/blob/master/CHANGELOG.md#security

PR #124, opened 2015-09-07
Update README.md

joepie91:
It's too easy to skim the documentation and accidentally use `jwt.decode` instead of `jwt.verify`. This adds a warning to prevent that.


jfromaniello:
Thanks!

PR #125, opened 2015-09-09
Fix typo in docs to correctly show asymmetric

cstaton:
I believe there is a typo in the docs that should say asymmetric instead of symmetric. This potentially could be confusing to readers. I have attached a picture with a red box around what I believe should be changed.

Thanks for the awesome library!

docstypo


FrankHassanabad:
Isn't symmetric the correct word?  It looks to be HMAC which is symmetric not asymmetric

jfromaniello:
This is symmetric indeed.

Issue #106, opened 2015-07-08
Question: sign Async method


kikar:
How come there is no Async method for sign or decode?

flore2003:
I think this is a valid issue. [node-jws](https://github.com/brianloveswords/node-jws) has a `createSign` method that provides async signing. It should be just a matter of using this, if the `sign` method is passed a callback.

PR #127, opened 2015-09-27
added async signing

lukewestby:
This PR addresses issue #106. Feedback would be appreciated, especially around whether the unit test is sufficient. Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,7 +35,7 @@ JWT.decode = function (jwt, options) {
   return payload;
 };
 
-JWT.sign = function(payload, secretOrPrivateKey, options) {
+JWT.sign = function(payload, secretOrPrivateKey, options, callback) {
   options = options || {};
 
   var header = {};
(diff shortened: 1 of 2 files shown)


syzer:
:+1:

jfromaniello:
Thank you! pulished as `v5.1.0`

syzer:
awesome!!! :+1:

lukewestby:
No problem! This library is a cornerstone for me, I'm happy to help.

Nepoxx:
FYI: While this is great, it does not respect the error first callback convention. This unfortunately means that for instanct this does not work with Bluebird's `promisify` or `.fromCallback()`.

lukewestby:
@Nepoxx since this function doesn't have a failure case would you suggest just always passing `undefined` to the first arg of the callback?

Nepoxx:
I'd say `null`.

This isn't big issue for me, changing this would break backwards compatibility. I **think** it should be changed to return `null`/`undefined` as to be more standard (`verify`'s callback uses error-first).

Thoughts?

PR #144, opened 2015-12-01
Add payload object literal warn to JWT.sign

KAndersonUS:
Checking the constructor of payload on sign and warning if it's not an object literal.

Helpful for those who tend to only read the code snippets in README.md

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,6 +44,10 @@ JWT.sign = function(payload, secretOrPrivateKey, options, callback) {
     header.typ = 'JWT';
   }
 
+  if (payload.constructor !== Object && options.keys.length) {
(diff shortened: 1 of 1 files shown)


jfromaniello:
I think this is on the right direction to avoid confusion but I will rather throw an error when the payload is string and the user specify an expiresIn, audience or subject

KAndersonUS:
Thanks, José! That would work too. However, I ran into confusion when passing in mongoose Model objects. ExpiresIn, audience, etc. didn't persist with those objects. Is there a case for not checking for Object literals as I'm proposing?

jfromaniello:
Can you try again on latest version? I think I found the issue with mongoose. Thanks

PR #146, opened 2015-12-05
add clock drift tolerance

jacopofar:
When I use the library to generate a token which is validated by different machines with a time different by a few seconds, it results as invalid.

This change adds the `ClockTolerance` field, and relative documentation and tests, to allow developers to add some tolerance in this scenario.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ encoded private key for RSA and ECDSA.
 * `issuer`
 * `noTimestamp`
 * `headers`
+* `clockTolerance`
 
 If `payload` is not a buffer or a string, it will be coerced into a string
 using `JSON.stringify`.
@@ -40,6 +41,8 @@ If any `expiresIn`, `audience`, `subject`, `issuer` are not provided, there is n
 Additional headers can be provided via the `headers` object.
 
 Generated jwts will include an `iat` claim by default unless `noTimestamp` is specified.
+The `iat` will be backdated of the amount of seconds specified in `clockTolerance`, if defined. Use this field when the token has to be validated by a machine which may have a slightly different system time due to clock drift.
+
 
 Example
 
--- a/index.js
+++ b/index.js
@@ -53,15 +53,14 @@ JWT.sign = function(payload, secretOrPrivateKey, options, callback) {
     });
   }
 
-
-var clockTolerance = 0;
-if (options.clockTolerance) {
-    if (typeof options.clockTolerance === 'number' && options.clockTolerance >= 0) {
-      clockTolerance = options.clockTolerance;
-    } else {
-      throw new Error('"clockTolerance" should be a positive number of milliseconds');
-    }
-}
+  var clockTolerance = 0;
+  if (options.clockTolerance) {
+      if (typeof options.clockTolerance === 'number' && options.clockTolerance >= 0) {
+        clockTolerance = options.clockTolerance;
+      } else {
+        throw new Error('"clockTolerance" should be a positive number of seconds');
+      }
+  }
   var timestamp = Math.floor(Date.now() / 1000);
   if (!options.noTimestamp) {
     payload.iat = payload.iat || timestamp - clockTolerance;
--- a/index.js
+++ b/index.js
@@ -53,9 +53,18 @@ JWT.sign = function(payload, secretOrPrivateKey, options, callback) {
     });
   }
 
+
+var clockTolerance = 0;
+if (options.clockTolerance) {
+    if (typeof options.clockTolerance === 'number' && options.clockTolerance >= 0) {
+      clockTolerance = options.clockTolerance;
+    } else {
+      throw new Error('"clockTolerance" should be a positive number of milliseconds');
+    }
+}
   var timestamp = Math.floor(Date.now() / 1000);
   if (!options.noTimestamp) {
-    payload.iat = payload.iat || timestamp;
+    payload.iat = payload.iat || timestamp - clockTolerance;
   }
 
   if (options.expiresInSeconds || options.expiresInMinutes) {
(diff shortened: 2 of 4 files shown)


jfromaniello:
I will rather have an option for `issuedAt`

jacopofar:
You mean having the `issuedAt` option instead, which would be a numeric date and overwrite the `iat` value? I chose this naming to discourage an usage different from introducing a small clock tolerance, would there be a different use case?

cvillerm:
I like the idea of this additional field, but I wasn't thinking at the same implementation for the proposed fix:

`payload.iat = payload.iat || timestamp - clockTolerance;`

In case there an `iat` in the payload, it should be used. This statement was only supposed to use a default `iat` timestamp if none was in the payload.
I rather think that the `clockTolerance` should be used in the timestamp comparison code:

`if (payload.nbf > Math.floor(Date.now() / 1000))`

and note by the way that the field to adjust for a tolerance is `nbf`, so the check should become:

`if (payload.nbf > Math.floor(Date.now() / 1000) - clockTolerance)`

Do you agree with it?

jacopofar:
I didn't consider the `nbf` field, sounds a good idea. So, when `nbf` is defined and is lower than `iat`, a JWT can be "from the future" and still considered valid considering the `nbf` by the various JWT libraries around? In that case I should change the code and move the option in the `verify` part.

I see that to overwrite the `iat` field is sufficient to place the value in the payload, so the library already can sign a token using an arbitrary issue timestamp.

cvillerm:
I am realizing that we may not talk about the same thing. We can (should IMHO) adjust for clock skew when creating such timestamp claims and when validating them. Adjustment can be done on both sides. I was referring to the latter but both can be useful as we may not control both the creating and validating side. 

First of all, an `iat` claim is mostly informational and isn't used as such in validations. I don't think that we should adjust it.

The `nbf` and `exp` claims are used for validation (if they are available). For short-live token the `exp` claim is usually defined based on a duration interval and this may already be enough to take into account clock skew. Adjusting the `nbf` claim for clock skew is what we should care the most about and I think that we agree on it, especially as this may often be based on the current time (likely to be what is and should be in `iat`) and clock skew can make validation in the next seconds fail.
We may think at adjusting it both when creating such a claim and when validating it. Based on that, there would be 2 places where `ClockTolerance` should be used.
Let me know what you think of it.

jacopofar:
The user case I'm dealing with is an application emitting a JWT using this library, which is distributed and verified by many other applications using other libraries in different servers with a different clock (having an error of about 6 seconds), and considered invalid until the clock is fixed.
I was looking for a maintainable way to introduce a tolerance for this scenario, working completely on the signing side which is the only I can work on, but couldn't find any so made this PR.

Now I've seen that actually it was already possible by putting a slightly backdated `iat` in the payload, not in the options. At this point I think this change is not necessary anymore, I suggest these changes:

* documenting in the README that the `iat` can be put in the payload
* introduce a check for the `iat` to be a number
* introduce the `clockTolerance` field in the `verify`

if you agree I'd proceed with updating the PR branch with these changes.

cvillerm:
Note first that I don't have authority to approve or not such or such change. I just discovered this plugin a few days ago and found your suggestion interesting and required IMHO, hence my comments to support it.
My use case is currently for the verify side and adding a `clockTolerance` (when checking the validity of `nbf`) is the only way to make the end to end JWT work.

About your use case, I still think that adding and using tolerance in the `nbf`field should be what to implement. You are right that you have control over it and that this wouldn't necessarily require an update of the plugin. The `nbf` claim is what the verification side should be checking and hopefully it is. I don't think that the `iat` should be adjusted. It should represent the issuing time. It would be good to check on your side if only adjusting (or setting manually) this claim would make your use case work.

I also hope that somebody who has authority to approve this change could comment about it.

jfromaniello:
I agree with this:

> - documenting in the README that the iat can be put in the payload
> - introduce a check for the iat to be a number
> - introduce the clockTolerance field in the verify

jfromaniello:
Thanks and sorry for the delay.

lmillott:
Hi guys,
I was searching for a way to allow for clock skew in node.js when validating jwt and came across this thread. I agree with @jfromaniello 's comment that it should be done in the verify side. For my scenario the token is issued by another service and we don't control how they generate the nbf time. We need to introduce the tolerance for skew on the validation side. Any chance we can revive the PR or start a new one for this issue?

cvillerm:
In the meantime, I found option `ignoreNotBefore` that can be good enough in case tokens created are valid as soon as they are created. This will skip checking the `nbf` claim.

lmillott:
Thanks @cvillerm!
My interpretation of the jwt spec is that the nbf claim must be used if it is provided. 
"The processing of the nbf claim requires that the current date/time MUST be after or equal to the not-before date/time listed in the nbf claim. Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew."

Ignoring it is an OK workaround for now, but I think adding the clock skew tolerance is a really important feature so that clients can do proper jwt validation.

jacopofar:
@lmillott  I think in one week I'll do another fork + PR with the result of this discussion (that is, state in the README that the you can put an `iat` in the payload, add an option to the verify part to tolerate the Clock Skew), if you are interested. My use case is the opposite (I generate the token, and an application outside my control using another library validates it on another server with a different clock), but it's nice to have a complete coverage.

lmillott:
@jacopofar That's great, thank you for contributing! Can't wait!

jacopofar:
Here you go https://github.com/auth0/node-jsonwebtoken/pull/188

lmillott:
Awesome!

Issue #111, opened 2015-07-16
Return a Promise


xpepermint:
Methods should return a Promise.

akohout:
+1

omsmith:
Everything that occurs during signing/verifying/decoding is completely synchronous. What benefit would a promise provide?

jden:
Verify currently offers a synchronous operation or an optional callback (which simply defers evaluation to a future turn of the event loop). It's unclear to me why the callback version exists. I also think it's an antipattern to overload the same method to behave both synchronously and asynchronously depending on the arguments.

xpepermint:
What I wanted to say is that `verify` should return a Promise :).

omsmith:
Sure, though as @jden pointed out the callback version of `verify` is essentially useless. The whole operation is synchronous. The callback, or a Promise, seem to be strictly detrimental in this case.

xpepermint:
Well... as I can see the sync version of `verify` will throw an error which means you have to put it inside the try/catch block. I prefer using promises instead of try/catch block because ~70% of my code is async (promises). If I could do verify.then(()=>something) it would look more elegant.

@omsmith I agree that the callback is useless and I would be more pleased if the sync version would just return `null` in case of an error (but you have different errors). I don't know what's best...

akohout:
As already pointed by @xpepermint, using promises would be more consistent with the rest of the code.

jden:
Could any of the maintainers comment about whether the intention is to refactor the implementation of `.verify` to use async versions of crypto functions?

I'm still +1 to separating async/sync into separate functions (as node core does, e.g. in `fs`), and +0 on callbacks vs promises

omsmith:
While streaming would be possible, there is no async sign/verify in crypto

jden:
@omsmith good point, and at the size that JWTs are, streaming is unnecessary overhead. My vote would be to remove the async function signature then.

dschenkelman:
Hey everyone,

There are different things here.

* Returning a promise - I don't think we would want to do that. If we accept callbacks anyone can wrap that in a promise, either using `new Promise(function(...){}` or something like bluebird's `promisifyAll`. We like promises (and use them a lot internally) but we don't think it is a good idea to expose them in APIs when they can be so easily wrapped. Some libs also both return a promise and a callback, but that leads to being tied to an implementation of promises which is not always desired.

[...]

PR #165, opened 2016-01-17
always use jws.sign such that errors can be at all handled

omsmith:
This PR switches `sign` to always use `jws#sign` instead of the streaming interface. Why?

1. I was going to open a PR to handle `error` events from the stream, however `callback` is not an _errback_ (oops!), as such the only way to actually catch/handle errors if for them to throw synchronously - and thus we need to use the more-synchronous method.
2. The stream interface calls `jws#sign` anyway, after buffering input data. In this case, all inputs are already known (and small), meaning the streaming interface just adds overhead
3. Similar to 2, and my sentiment in #111, this is all synchronous anyway, and the callback APIs provide little-to-no value (though atleast `verify` is an errback!)

### DIFF
--- a/index.js
+++ b/index.js
@@ -127,15 +127,20 @@ JWT.sign = function(payload, secretOrPrivateKey, options, callback) {
     encoding = options.encoding;
   }
 
-  if(typeof callback === 'function') {
-    jws.createSign({
-      header: header,
-      privateKey: secretOrPrivateKey,
-      payload: JSON.stringify(payload)
-    }).on('done', callback);
-  } else {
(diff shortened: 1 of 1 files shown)


jfromaniello:
You are right... But I'd rather change the next major version to do be an errback

```
jws.createSign({
   header: header,
   privateKey: secretOrPrivateKey,
   payload: JSON.stringify(payload)
}).once('done', function (signature) { callback(null, signature); } )
  .once('error', callback); 
````

does it make sense?

jfromaniello:
I will close this in favor of handling #169 in the next major release.

Issue #122, opened 2015-08-26
Refresh token


matiux:
Any ideas of how to implement the refresh token?

Seems nobody has written nothing...

jfromaniello:
http://stackoverflow.com/questions/26739167/jwt-json-web-token-automatic-prolongation-of-expiration

matiux:
OK thanks.

But some example of how to implement with node-jsonwebtoken?

Sean-Wang:
+1

nnur:
+1

nnur:
is there a way to reissue the same token but just change the expiry date?

morgondag:
kinda same as #133 but is there any official support to refresh a token?
Right now I validate the token and user data and then sends a new one to the client if it has expired, 
but would rather just update the existing token.

```js
    jwt.updateToken(token, data)
```

mateeyow:
@morgondag I think your implementation is very dangerous since you are not assigning any expiration time to the token. It means that if some hacker gets the token, he/she can access your API indefinitely.

morgondag:
@mateeyow sounds good, so what we are asking for is a way to update a token. 
nothing is saying that we are nit doing any form of validation on it beforehand. 

So there is no real way to update a token or change it expiry date without generating a new key. 

I guess the best way is simply to force to user to login again and again and again :-/

alexbooker:
I am also having trouble finding good information on:

- How to implement refresh tokens server-side
- A sensible strategy to refresh tokens on the client-side

varunvairavan:
+1

morgondag:
any progress?

umair321:
you can verify user using previous token, then generate a new one for that user. Like:
 jsonwebtoken.verify(token, mySecret, function (err, user) {
            if (err) {
                return res.json(err);
            }
            if (user.id)
              res.json({
newToken: jwToken.issue({
                        id: user.id
                    })
});
           
        });

jppellerin:
This is what I ended up doing on my end.  My token is already verified by the time I hit my refresh function.

```
static refreshToken = (token): string => {
    let optionKeys = ['iat', 'exp', 'iss', 'sub'];
    let newToken;
    let obj = {};

    let now = Math.floor(Date.now()/1000);
    let timeToExpire = (token['exp'] - now);

    if (timeToExpire < (60 * 60)) { //1h
        for (let key in token) {
            if (optionKeys.indexOf(key) === -1) {
[...]

PR #172, opened 2016-02-03
Added API endpoint for token refresh

jppellerin:
According to issue #122, there was some interest in having an endpoint to refresh a token.

A refresh is considered to have the same token returned, but with a later expiry time.

This function takes in a decoded token, copies the attributes of this token over, and with that data, proceeds to signing a new token.  The only piece of data that is replaced is the `expiresIn` value, which is passed in the function parameters.  By extension, the `iat` value is also changed since a new token is created.


Kiura:
+1

foobar8675:
+1

jfromaniello:
Very nice job, I will merge this!

jppellerin:
:+1: Excellent!  Thank you!

mateeyow:
Quick question, when will this be merged? Any ETA?

pmeijer:
+1

arturskeeled:
+1 Any news on when is this being implemented?

jppellerin:
@mateeyow @arturskeeled I'm not sure when the next major release will be coming out, but I have just updated my forked version with the refresh endpoint : https://github.com/jppellerin/node-jsonwebtoken.  You could use this for development purposes.

*_Disclaimer_*: I have not tested the merge thoroughly..

rickli1989:
Waiting for this to get released

rickli1989:
When is the next major release for this ?

RWOverdijk:
👍  we'd really like to implement refresh tokens.

VMBindraban:
Any ETA yet?

RWOverdijk:
Bump...

RWOverdijk:
Last bump. I don't want to fork but I really need this.

jppellerin:
@RWOverdijk : I have just updated my fork with the newest version and fixed the tests.
https://github.com/jppellerin/node-jsonwebtoken

I can keep this up to date with the newest features here until it gets merged in.  If you want some assurances, I can tag you a version.

That's the best I can do.

lendi-au-glen:
Would be really great to have this in. 💃

README.md
```diff
@@ -172,6 +172,28 @@ console.log(decoded.header);
 console.log(decoded.payload)
```
Tchikita35: Good

README.md
```diff
@@ -172,6 +172,28 @@ console.log(decoded.header);
...
 ```
 
+### jwt.refresh(token, expiresIn, secretOrPrivateKey [, callback])
+
+Will refresh the given token.  The token is __expected__ to be *decoded* and *valid*. No checks will be performed on the token.  The function will copy the values of the token, give it a new expiry time based on the given `expiresIn` parameter and will return a new signed token using the `sign` function and given secretOrPrivateKey.
```
Tchikita35: ⭐⭐⭐⭐⭐

mitchellporter:
Is there a reason this hasn't been merged yet?

morgondag:
forking is the place.

Isaac6702:
Now you can use this?

jfromaniello:
I am still considering this feature, I think as it is it doesn't add too much value over `sign` and it creates another api to maintain and another source for confusion and vulnerabilities. I `refresh` should also validate the token it would be more like a verify->sign.

jppellerin:
@jfromaniello Thanks for the feedback.  The goal (for what we are using it for) is to extract all fields of a token and return it with a new `exp`.  The only way (as far as I know) is to create a new one.

I didn't include verification since the token needs to be decoded for this to work.  Therefore, it assumes the user has manipulated this - thus it would be safe to assume that it has been validated.

I agree that if one uses this API the wrong way and refreshes a non-valid JWT, then a valid one will be returned, which has cause for concern.

The value this endpoint has is to save the logic of copying over the fields of the existing token and re-signing it.  I can agree with you that it is not something extremely difficult to do, but is something that is useful and that many people are using (or so it seems by the interest of this pull request).  I would assume that this is going to be more and more common as people are moving towards serverless architectures/completely stateless.

All that being said: I think that verification would be a good idea.  I also think it is a useful endpoint (I may be biased since I'm using this).

README.md
```diff
@@ -172,6 +172,28 @@ console.log(decoded.header);
...
+* `secretOrPrivateKey` : is a string or buffer containing either the secret for HMAC algorithms, or the PEM
+encoded private key for RSA and ECDSA.
+* `callback` : If a callback is supplied, callback is called with the newly refreshed JsonWebToken string
+
+Example
+
+```js
+// ...
+var originalDecoded = jwt.decode(token, {complete: true});
+var refreshed = jwt.refresh(originalDecoded, 3600, secret);
```
AlexanVal: I think it should be `var refreshed = jwt.refresh(**token**, 3600, secret);`

ForsakenHarmony: ?

jppellerin:
@jfromaniello Maybe this pull request should be rejected if it's not going to be merged?  It has been over a year now.

fiftysoft:
@jppellerin can you give me an example to use it ?

ndaljr:
Whats the update on this pull request? It's been over a year. I agree with @jppellerin that it should either get merged if not rejected. @jppellerin Have you considered making a fork of this and pushing to npm?

jppellerin:
@fiftysoft : An example is in the readme docs of the pull request.

@ndaljr : I have considered making a fork, but the problem is keeping up with the rest of the library.  I don't want to have the code duplicated and potentially falling behind on updates.  If this was to become it's own module, it would have to only extend it.  Which, I might put some thought into this.

oshalygin:
Any status on this?

README.md
fiddur: I don't see the point of this.  You have to decode and validate it, and you will have the payload, why not just call sign on that payload instead of having a refresh function that will decode it again?

fiddur:
Fixed by documentation: #371

VMBindraban:
Well this feels like a big slap in the face for the people waiting for it.

fiddur:
@VMBindraban: Since you still had to decode and validate it before using the refresh function, that function added very little value over just deleting old exp/iat and using sign instead.

Keeping the API and the code clean and small without unnecessary complexity is more important than a convenience method that is hardly even convenient.

BrOrlandi:
I have found a problem in the refresh function that's causing it to generate the same token.
In the file [refresh.js line 102](https://github.com/jppellerin/node-jsonwebtoken/blob/5f11be4d37b1c644cb01fbb48b67af33d7899c84/refresh.js#L102)
the `token.iat` is trying to read a property that doesn't exist because `token` is a string.
Even using `payload.iat`, the code allows passing the `iat` in payload, which causes the `jwt.sign()` generates the same token. The solution I found is to remove the `iat` from payload. Anyone had this problem?

deepakchandola717:
What if the original token gets stolen?

RWOverdijk:
Then you call the police. Theft is **not** okay.

image

deepakchandola717:
I mean this feature could pose a security issue if any app is using it. There's no point in making it expire only in first place.

RWOverdijk:
You're commenting on an issue from 2016 dude.

But to answer your question: every form of session/auth has the same issue. Cookies: get the token. JWT: get the token. oauth: get the token.

Just make it really really difficult to steal the token. 😄

deepakchandola717:
Ok you just think why do we set an expiry to the token in first place and why do we have two different types of token.
Make it really really really difficult to steal, never let it expire and hope you did everything right 😂

deepakchandola717:
This is a very old issue and every one was asking why feature wasn't merged. Thought this could be a reason.

PR #188, opened 2016-03-24
Origin/clock skew tolerance

jacopofar:
As discussed in a [previous PR](https://github.com/auth0/node-jsonwebtoken/pull/146), three changes:

-     documenting in the README that the iat can be put _in the payload_
-     introduce a check for the iat to be a number
-     introduce the clockTolerance field in the verify function


jfromaniello:
Thank you very much for working on this and sorry for the big delay

PR #189, opened 2016-03-25
Fix verification of unsigned tokens (fix #185)

### LINKED ISSUE #185: Not possible to verify unsigned token (signed with alg none)
Is it possible to verify (as in `jwt.verify()`) a token that was signed with algorithm `none`?

```javascript
var payload =  {
    iss: 'https://example.com'
    aud: 'test.client.id'
};
var signedToken = jwt.sign(payload, null, { algorithm: 'none' });

// First try with empty key, causes 'secret or public key must be provided'
jwt.verify(signedToken, null, { audience: 'test.client.id', issuer: 'https://example.com' }, ...);

// Second try with a dummy key, causes 'jwt signature is required'
jwt.verify(signedToken, '<dummy key>', { audience: 'test.client.id', issuer: 'https://example.com' }, ...);
```

I expected the first version to do what I want, which is to realize that the token does not have a signature and therefore doesn't require a signature validation and hence not a key either.

jonekdahl:
`verify()` now accepts null for secretOrPublicKey if the token does not contain a signature. The default algorithm for unsigned tokens is `none`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -172,12 +172,18 @@ JWT.verify = function(jwtString, secretOrPublicKey, options, callback) {
     return done(new JsonWebTokenError('jwt malformed'));
   }
 
-  if (parts[2].trim() === '' && secretOrPublicKey){
+  var hasSignature = parts[2].trim() !== '';
+
+  if (!hasSignature && secretOrPublicKey){
     return done(new JsonWebTokenError('jwt signature is required'));
   }
 
-  if (!secretOrPublicKey) {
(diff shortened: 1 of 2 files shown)


jonekdahl:
I made a small improvement to the order in which the jwt is checked. If there are suggestions for further improvements, I'd be happy to address them so please let me know.

jfromaniello:
released as `v6.1.0`

jfromaniello:
Thank you very much and sorry for the delay

PR #192, opened 2016-04-27
feat: change .sign to standard async callback

popomore:
- subscribe error event
- change the first argument of the callback as error

### DIFF
--- a/index.js
+++ b/index.js
@@ -132,7 +132,11 @@ JWT.sign = function(payload, secretOrPrivateKey, options, callback) {
       header: header,
       privateKey: secretOrPrivateKey,
       payload: JSON.stringify(payload)
-    }).on('done', callback);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -132,7 +132,11 @@ JWT.sign = function(payload, secretOrPrivateKey, options, callback) {
       header: header,
       privateKey: secretOrPrivateKey,
       payload: JSON.stringify(payload)
-    }).on('done', callback);
+    })
+    .on('error', callback)
```
popomore: https://github.com/brianloveswords/node-jws/blob/master/lib/sign-stream.js#L62

popomore:
It's a breaking change

NOtherDev:
+1
Docs update also needed.

PR #219, opened 2016-06-22
Exp calculated based on iat. fix #217

### LINKED ISSUE #217: exp time not being properly calculated if setting iat
The expected behavior as per the documentation is that if you "iat" flag is populated, it will calculate the "exp" time off of that, instead of the current time. This is not the case.

`payload.exp = timespan(options.expiresIn);`

The timestamp function does not take into account for modified issued at time.

The function should be modified somehow in this manner:

```
 module.exports = function (time, iat) {
   var timestamp = iat || Math.floor(Date.now() / 1000);

   if (typeof time === 'string') {
     var milliseconds = ms(time);
     if (typeof milliseconds === 'undefined') {
      return;
     }
     return Math.floor(timestamp + milliseconds / 1000);
   } else if (typeof time === 'number' ) {
     return timestamp + time;
   } else {
     return;
   }

 };
```

This would allow for passing in of the starting time so that the exp prop can be properly calculated instead of always calculating it based off the current time.

This issue arose when writing a test to verify an expired JWT and I back dated the issued at time by 20 minutes and set the expiresIn signing option for "15m".
I saw a valid "iat" of 20 minutes ago, but the "exp" to be 15 minutes from the current time, thus offering a "JWT" that was valid for 35 minutes.

lusarz:


jfromaniello:
Thank you very much and sorry for the delay, released as `v7.1.0`.

PR #231, opened 2016-07-21
Use lodash.once instead of unlicensed/unmaintained cb

holm:
I am trying to get all our dependencies (including transient) to have proper license info, so we can automate the checking. We found that 'cb' doesn't have a license set (in the published version), and haven't seen any activity for over 3 years.

This switches to use lodash.once instead, which should be as small and as efficient.


jfromaniello:
Thanks a lot!

PR #232, opened 2016-07-22
Edit readme to include caution about excessive payload size

sdotson:
This might be a rookie mistake, but I got the initial /authenticate route working, and it returned the correct token, but on every subsequent request, the payload information was being lost.

I eventually figured out that it had something to do with the size of the payload I was including. I was including all the user properties in the payload. If you limit it to one or two, things started to work!

I'd like to update the readme to mention this limit. Might save some others some grief!

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ $ npm install jsonwebtoken
 
 (Synchronous) Returns the JsonWebToken as string
 
-`payload` could be an object literal, buffer or string. *Please note that* `exp` is only set if the payload is an object literal.
+`payload` could be an object literal, buffer or string. *Please note that* `exp` is only set if the payload is an object literal. Payloads that are excessively large will cause errors.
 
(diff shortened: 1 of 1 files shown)


jfromaniello:
Are you sure the issue was with the size? Maybe it was due to a circular dependency:

```
> var a = {}
undefined
> a.a = a
{ a: [Circular] }
> JSON.stringify(a)
TypeError: Converting circular structure to JSON
    at Object.stringify (native)
    at repl:1:6
    at REPLServer.defaultEval (repl.js:262:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:431:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:211:10)
    at REPLServer.Interface._line (readline.js:550:8)
```


Anyway there is another issue where we are discussing adding a default limit to the payload, but I would like to also know where it throws.

Thanks!

PR #235, opened 2016-08-10
Removing unnecessary extra decoding. jwtString is already verified as…

venatir:
jwtString was decoded twice. There was no need for that. This becomes quite slow for huge strings.


venatir:
Please ignore. I fixed this and will make another pull request

PR #257, opened 2016-10-13
Add the expirationRequired option

### LINKED ISSUE #220: can't check that a JWT has an expiry defined in the verify function
On the verify function, I can set the verification options to check the issuer, audience etc.  These trigger a verification failure if the field is either wrong or completely missing.  But there is no way to insist that there is an expiry in the claims.

I thought that I could set `ignoreExpiration` to `false` and force the verification to check, but it seems that the `ignoreExpiration` option only has an effect (and the expiry is only checked) if the exp field exists in the claims.

oliverdunk:
Adds the expirationRequired option to the verify method. When this is present and set to true, the exp header must exist or the verification will fail immediately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ encoded public key for RSA and ECDSA.
 * `audience`: if you want to check audience (`aud`), provide a value here
 * `issuer` (optional): string or array of strings of valid values for the `iss` field.
 * `ignoreExpiration`: if `true` do not validate the expiration of the token.
+* `expirationRequired` if `true`, an expiration must be specified. May become default in the future.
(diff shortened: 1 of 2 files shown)


oliverdunk:
A lot of checks seem to have failed considering the small amount of code changed - anyone have any ideas?

jcw-:
My guess - the original code was guarded by the `typeof payload.exp !== 'undefined'` check, your change allows a flow into it now under the conditions of `options.ignoreExpiration == false` and `options.expirationRequired == false`.

oliverdunk:
Bump?

panva:
Closing this as stale.

PR #258, opened 2016-10-14
Misspelled algorithm name

LeonardLaszlo:


jfromaniello:
It is correct. RS256 is a different algorhitm, main different is that is "asymmetric". This means that you sign with a private key , and validate with a public key.

HS256 is symmetric, you sign and verify with a single secret.

In the example, the variable name is called `cert`. So, I think is correct.

LeonardLaszlo:
You are right, sorry for that!

Issue #85, opened 2015-05-06
Documentation uses payload when it should be claim


samholmes:
The JWT spec calls this _payload_ a _claim_. You're documentation should use this convention to not confuse the subject.

jonrh:
Although I find *payload* to be a more descriptive word I agree that *claim* should be used for consistency with the spec. Two possible ways to address this:

* Refactor *payload* to *claim* in code and documentation.
* Add to the top of repository README.md: 
	* *payload is used interchangeabliy with claim*

panva:
The `payload` would be the JWS payload, in JWT case a JSON formatted string with attributes being individual `claims`. I don't think the documentation needs updating. If anything than `payload` could be `claims`, but that becomes too darn hard to keep track of when reading - `claim(s)` everywhere, whats what?

PR #259, opened 2016-10-26
added payload can be used interchangeably with claim

303Rickp:
added: 'payload can be used interchangeably with claim'


Dunbarry:
Related to issue #85

panva:
See https://github.com/auth0/node-jsonwebtoken/issues/85#issuecomment-505423691

Issue #48, opened 2015-01-20
Error on insufficient key length


MichielDeMey:
Might I suggest we throw an error when the user tries to sign a token with a key of a length with an insufficient length?

I believe a length > 8 should be sufficient against brute-force attacks.

Let me know what you guys think.

jfromaniello:
Maybe `console.warn` instead? I'm not sure

MichielDeMey:
That definitely sounds more reasonable, throwing an error would be too aggressive.

panva:
Closing this as stale. Using sufficient entropy secrets and higher bitsized RSA keys remains the responsibility of the user, not of the library to enforce or log about. We could make a recommendation in the README though, possibly with a link to the [JWT BCP](https://tools.ietf.org/html/draft-ietf-oauth-jwt-bcp-06#section-3.5)

PR #260, opened 2016-10-26
key length error

stephen-knutter:


catiecook:
This is referring to issue #48

panva:
Same as https://github.com/auth0/node-jsonwebtoken/issues/48#issuecomment-505420109

PR #261, opened 2016-10-26
added the check for at least 8 characters with a warning

catiecook:
referencing issue #48 and continuing pull request #260 to pass test


jfromaniello:
sorry for the long delay, maybe we should use  `should` instead of `must`

panva:
Same as https://github.com/auth0/node-jsonwebtoken/issues/48#issuecomment-505420109

PR #272, opened 2016-11-17
Bump node-jws version number (Buffer DeprecationWarning)

oconn:
Bump node-jws version number from ^3.1.3 -> ^3.1.4. Version 3.1.3 has a dependency on base64url ~1.0.4 which is throwing the NodeJS Buffer DeprecationWarning. version 3.1.4 pulls in base64url ^2.0.0 which address the issue. With `"jws": "^3.1.3"` specified, jsonwebtoken will pull the correct version (3.1.4), bumping the minimum version to ^3.1.4 will fix caching issues.


jfromaniello:
Thanks and sorry for the delay, published as `v7.1.10`

PR #285, opened 2016-12-21
Remove callback API

### LINKED ISSUE #246: Deprecate callback API

freitagbr:
Fixes #246 

Deprecates the callback API, leaving only the synchronous API. The methods exposed by this package are synchronous internally, so there is no benefit in making them asynchronous.


graingert:
@ziluvatar wasn't this going to go into 8.0.0?

ziluvatar:
I decided to leave it out for now and package the other breaking changes alone. I am still thinking about this change and the suffering we will generate for people to migrate (easy change from our point of view, but each consumer has its own business). The same way the callback api is useless on verify (nowadays) removing it gives more pain that benefit to current consumers.

I also have a pending PR where having a callback will make you possible to add some verifications that could require asynchronous operations.

Summary: leaving the API with callback remove possible migration pain and leaves open the door for future async verifications. Nevertheless I don't close this issue because it's still something we could apply (not decided yet).

ziluvatar:
Finally we took advantage of the callback on `verify` that will behave async if needed by the consumer.
See: https://github.com/auth0/node-jsonwebtoken/pull/480

I'm closing this PR. Thanks @freitagbr for the effort even if it did not get merged.

PR #292, opened 2017-01-05
Upgrade joi to v10.1.0

chinclubi:
Moment.js was removed from joi since v10.0.0 .... please remove "BIG" module from dependencies. 💃

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "url": "https://github.com/auth0/node-jsonwebtoken/issues"
   },
   "dependencies": {
-    "joi": "^6.10.1",
+    "joi": "^10.1.0",
     "jws": "^3.1.4",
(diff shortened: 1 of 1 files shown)


ziluvatar:
Similar to https://github.com/auth0/node-jsonwebtoken/pull/301

> After a some investigation, any upgrade of joi from the current version will break compatibility with ES5.
https://github.com/hapijs/joi/commits/master?after=Y3Vyc29yOtCeNCeSeSsaqQE27cSdaZ5UG57jKzI0NA%3D%3D

> Let's keep it as is for now, I'll mark it as next major (although I'm not sure if it will be the next one or the following one)

ziluvatar:
Closing this in favor of: https://github.com/auth0/node-jsonwebtoken/pull/348

PR #294, opened 2017-01-11
Raise jws.decode error to avoid confusion with "invalid token" error 

evolvah:
jws.decode() never throws an error. At least, in its current version. However, if it were to throw an exception, the diagnostics would be indistinguishable from a soft failure to decode a token. I had an extra trailing space on my JWT and it took me some additional debugging work to trace the actual root cause because the error message was not distinct.

### DIFF
--- a/test/jwt.hs.tests.js
+++ b/test/jwt.hs.tests.js
@@ -77,6 +77,21 @@ describe('HS256', function() {
         done();
       });
     });
-
   });
+
+  describe('should fail verification graceflly with trailing space in the jwt', function() {
+    var secret = 'shhhhhh';
+    var token  = jwt.sign({ foo: 'bar' }, secret, { algorithm: 'HS256' });
+
+    it('should return the "invalid token" error', function(done) {
+      var malformedToken = token + ' '; // corrupt the token by adding a space
+      jwt.verify(malformedToken, secret, { algorithm: 'HS256', ignoreExpiration: true }, function(err, decoded) {
+        assert.isNotNull(err);
+        assert.equal('JsonWebTokenError', err.name);
+        assert.equal('invalid token', err.message);
+        done();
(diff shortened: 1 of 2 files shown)


ziluvatar:
Hi @evolvah ! What do you think to simply forward the unexpected error? We do that in other places for calling `jws`, at least we unify the way we treat them and we remove your concern as well.

If that's ok for you, can you modify your commit/PR message as well to reflect the new change?

evolvah:
That's a good idea! Please see the latest update to this PR, @ziluvatar.

evolvah:
I do like PRs with negative net LOC count :)

ziluvatar:
Me too!!! But it seems I expressed my suggestion badly :(
I meant to just forward the error:
```
  var decodedToken;
  try {
    decodedToken = jws.decode(jwtString);
  } catch(err) {
    return done(err);
  }
```

It is not negative net LOC count but some letters are removed ;)

evolvah:
Ah, I see now! Let me make the change and add a test case. Once I'm done, I will update the PR.

evolvah:
Done! With a new test case being added.

test/jwt.hs.tests.js
```diff
@@ -77,6 +77,21 @@ describe('HS256', function() {
         done();
       });
     });
+  });
+
+  describe('should fail verification graceflly with trailing space in the jwt', function() {
```
ziluvatar: typo, `gracefully`
Thanks for the new test :) 
Modify that and I'll merge.

evolvah:
Bummer! Fixed the typo, updated the PR. Ready to merge.

ziluvatar:
Thanks @evolvah 👏  , I modified the PR description to make more explicit the change.

PR #301, opened 2017-01-24
update joi dependency

wao813:
The joi version in project is too old.  Causing error for projects uses it with newer version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "url": "https://github.com/auth0/node-jsonwebtoken/issues"
   },
   "dependencies": {
-    "joi": "^6.10.1",
+    "joi": "^8.4.2",
     "jws": "^3.1.4",
     "lodash.once": "^4.0.0",
     "ms": "^0.7.1",


ziluvatar:
The v10 brought problems for ES5 compatibility https://github.com/auth0/node-jsonwebtoken/commit/51d4796c07344bf817687f7ccfeef78f00bf5b4f
Not sure about the v8 though.

zvictor:
right now, installing `jsonwebtoken` installs `joi`  indirectly and consequently `moment.js` because of hapijs/joi#985.

my app becomes significantly heavier for no reasonable reason.

ziluvatar:
After a some investigation, any upgrade of joi from the current version will break compatibility with ES5.
https://github.com/hapijs/joi/commits/master?after=Y3Vyc29yOtCeNCeSeSsaqQE27cSdaZ5UG57jKzI0NA%3D%3D

Let's keep it as is for now, I'll mark it as next major (although I'm not sure if it will be the next one or the following one)

ziluvatar:
Let's continue the possible conversation in [this other PR](https://github.com/auth0/node-jsonwebtoken/pull/292) (older) instead of having two PR for the same problem.

PR #303, opened 2017-01-31
rauchg/ms.js changed to zeit/ms

gswalden:
The ownership of this package has changed, see 301: https://github.com/rauchg/ms.js


ziluvatar:
Thanks @gswalden to realize about it! 👏

PR #305, opened 2017-02-01
Fix handling non string tokens

### LINKED ISSUE #304: TypeError when token is not a string
I started getting uncaught exceptions when receiving tokens that are not strings:

```
TypeError: jwtString.split is not a function
```

Since I'm only handling errors exposed by this library I expected a `JsonWebTokenError` to be thrown when verifying a value other than string.

I saw some issues related to this error but the focus was on throwing or return it on the callback. This way I suggest to test the type of the token before splitting it and throw a `JsonWebTokenError` if not a string.

ricardogama:
This PR adds handling of non string tokens, throwing a `JsonWebTokenError` error with `jwt malformed` message instead of `TypeError`.

Closes #304

### DIFF
--- a/test/issue_304.tests.js
+++ b/test/issue_304.tests.js
@@ -1,0 +1,41 @@
+var jwt = require('../index');
+var expect = require('chai').expect;
+
+describe('issue 304 - verifying values other than strings', function() {
+
+  it('should fail with numbers', function (done) {
+    jwt.verify(123, 'foo', function (err, decoded) {
+      expect(err.name).to.equal('JsonWebTokenError');
+      done();
+    });
+  });
+
+  it('should fail with objects', function (done) {
+    jwt.verify({ foo: 'bar' }, 'biz', function (err, decoded) {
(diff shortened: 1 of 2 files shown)


ricardogama:
@jfromaniello any change you can review this?

ziluvatar:
Hi @ricardogama, I see it ok, can you just change a silly thing? Instead of that error message, we can provide more information, like `jwt must be a string` or `jwt must be a string, but number was found`. What do you think?

ricardogama:
@ziluvatar Sure, I think there's no need to be so specific in the message, so I went forward with `jwt must be a string`. Sounds good?

ziluvatar:
Yes, it is enough, thanks @ricardogama !!

ricardogama:
@ziluvatar No problem at all! Can this fix be released?

ziluvatar:
@ricardogama we released v7.3.0 with this fix on it yesterday :)

ricardogama:
@ziluvatar Awesome, thanks!

PR #330, opened 2017-03-25
We don't need node 5, it's neither lts nor current

ForsakenHarmony:


ziluvatar:
@ForsakenHarmony We should only remove engines if we find a bug for that version where we don't want to / can't provide a solution. What do you think?

ForsakenHarmony:
hmm, true, but it's unnecessary testing

ziluvatar:
The only difference I can see is Travis running tests for node 5 if you have it in "engines" in the package.json

Master: https://travis-ci.org/auth0/node-jsonwebtoken/builds/209440400
This PR: https://travis-ci.org/auth0/node-jsonwebtoken/builds/214892125

I'd leave it as is if you don't mind.

PR #342, opened 2017-04-18
`expiresIn` and `notBefore` require milliseconds

ubergesundheit:
The Readme claims the keys `expiresIn` and `notBefore require either a number of seconds or a time span.
This commit corrects to milliseconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ encoded private key for RSA and ECDSA.
 `options`:
 
 * `algorithm` (default: `HS256`)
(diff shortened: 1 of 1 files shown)


ziluvatar:
They should be in seconds, since `exp` and `nbf` are expected in seconds. The string time span is just to have a nicer way to express the time.
Actually we have a [bug / enhancement](https://github.com/auth0/node-jsonwebtoken/issues/273) because `maxAge` is using milliseconds and it should use seconds.

Issue #343, opened 2017-04-18
Unnecessary dependencies 


prevostc:
Hi,

I've been using this lib with great success on AWS lambda.
My main issue is the generated package size which is quite important for a lambda function (faster code uploads + faster js parsing = less money spent) 

With a webpack builder and a default config, a simple lambda that only jwt.verify some token comes at a whopping 741.64 Kb 
- 113.93 Kb of Joi
- 62.78 Kb of isemail (dependency of joi ?)
- 455.55 Kb of moment (definitely a dependency of Joi)
- other small stuff

As I don't care about moment locales, I removed them with `new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/)`.

The final jsonwebtoken package size is 412.01 Kb 
- 113.93 Kb of Joi
- 62.78 Kb of isemail (dependency of joi ?)
- 221.42Kb of moment (definitely a dependency of joi)
- other small stuff

I do not know much about this project but it doesn't seem to need the full power of joi and moment. Joi is only used in sign.js and you could gain a lot by just getting rid of joi altogether.

That being said, it's just a suggestion because anyone could exclude unwanted dependencies with webpack, but it's a nice default to have.

jfromaniello:
I have had the same experience with Lambda so 👍 .

Sounds like we should get ride of Joi, use something smaller or maybe just manual validation.

borromeotlhs:
you guys should check the repo made for exactly this purpose in: https://github.com/hapijs/joi/issues/528

chrisprice:
@jfromaniello I've raised a PR (#348) which adds the manual validation. Again for me it's lambda performance which is driving this.

ziluvatar:
Solved in release 8.0.0

Issue #334, opened 2017-03-27
CSP Issue with downstream hapijs/joi dependency


jshcrowthe:
It seems that the listed version of the [hapijs/joi](https://github.com/hapijs/joi) in your `package.json` has a downstream dependency with a CSP issue. 

The Error being thrown is:
`EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".`

The stack trace points to this function in the nested version of [hapijs/isemail](https://github.com/hapijs/isemail/tree/v1.2.0):
https://github.com/hapijs/isemail/blob/v1.2.0/lib/isemail.js#L155-L172 

This fails CSP because of the string eval return at the end of the function.

This function (and from what I can tell, the CSP error) is removed in the next update of the library which was `v2.0.0` of `hapijs/isemail`.

This major update was released in `v7.0.0` of `hapijs/joi` (see commit [here](https://github.com/hapijs/joi/commit/341178b937948cde0bdfe2e8138ecf45200da1c6))

Full Disclosure: Haven't looked at all into how much work the major bump would be, or how many changes would be made on the node-jsonwebtoken side. 

All that said, is a version update possible `hapijs/joi`?

ziluvatar:
Currently we can not upgrade version without losing ES5 compatibility. We may do it for future major versions though. There are a couple of PRs asking for it.

SimonSchick:
Wouldn't replacing `joi` with a few equivalent asserts solve this problem?

ziluvatar:
Closing. JOI was replaced on v8.0.0

PR #348, opened 2017-05-09
Remove joi to shrink module size

chrisprice:
As noted in #343 & #292 the specific version of `joi` is bloating the package size. However upgrading to the latest version would break ES5 compatibility. 

This PR shrinks the module size from **3.12M** to **0.25M** by replacing joi with bespoke implementation using lodash helpers.

N.B. To keep the implementation as simple as possible, these changes mean that the validation error messages are marginally less helpful e.g. -

`"expiresIn" must be a number` versus `"expiresIn" is not the correct type`

This should also solve #334.

### DIFF
--- a/sign.js
+++ b/sign.js
@@ -11,23 +11,23 @@ var isString = require('lodash.isstring');
 var once = require('lodash.once');
 
 var sign_options_schema = {
-  expiresIn: function(value) { return isInteger(value) || isString(value); },
-  notBefore: function(value) { return isInteger(value) || isString(value); },
-  audience: function(value) { return isString(value) || isArray(value); },
-  algorithm: includes.bind(null, ['RS256', 'RS384', 'RS512', 'ES256', 'ES384', 'ES512', 'HS256', 'HS384', 'HS512', 'none']),
-  header: isPlainObject,
-  encoding: isString,
-  issuer: isString,
-  subject: isString,
-  jwtid: isString,
-  noTimestamp: isBoolean,
-  keyid: isString
+  expiresIn: { isValid: function(value) { return isInteger(value) || isString(value); }, message: '"expiresIn" should be a number of seconds or string representing a timespan' },
+  notBefore: { isValid: function(value) { return isInteger(value) || isString(value); }, message: '"notBefore" should be a number of seconds or string representing a timespan' },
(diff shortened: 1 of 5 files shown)


sign.js
```diff
@@ -1,29 +1,52 @@
...
+  nbf: isNumber
+};
 
+function validate(schema, unknown, object) {
+  if (!isPlainObject(object)) {
+    throw new Error('Expected object');
+  }
+  Object.keys(object)
+    .forEach(function(key) {
+      if (schema[key] == null) {
```
ziluvatar: `!schema[key]` ?

chrisprice:
Changes sound good, as soon as I get a chance I'll update the PR.

test/schema.tests.js
```diff
@@ -0,0 +1,135 @@
...
+    });
+
+    it('should validate algorithm', function () {
+      expect(function () {
+        sign({ algorithm: 'foo' });
+      }).to.throw(/"algorithm" must be a valid string enum value/);
+      sign({algorithm: 'RS256'});
+      sign({algorithm: 'RS384'});
+      sign({algorithm: 'RS512'});
+      // not sure how to fix error - 'Could not find expected "seq"'
```
chrisprice: Wasn't sure how to get these tests to pass, I assume I'm missing some required configuration for the `ES*` algorithms.

chrisprice:
So it took a little longer than I expect to get back around to doing this but I think I've made all of the suggested changes.

test/schema.tests.js
ziluvatar: You need to use different certificates for RS and ES. Probably it's that.
Take a look to: https://github.com/auth0/node-jsonwebtoken/blob/master/test/jwt.asymmetric_signing.tests.js#L13

test/iat.tests.js
```diff
@@ -12,11 +12,4 @@ describe('iat', function () {
     expect(result.exp).to.be.closeTo(iat + expiresIn, 0.2);
   });
 
-
-  it('should throw if iat is not a number', function () {
```
ziluvatar: Why do you remove it? is it because with this PR it is tested in the schema test?

chrisprice: Yep, that's correct.

chrisprice:
Thanks for the pointer on ECDSA signing, I've updated the tests to use the other key where appropriate.

Issue #346, opened 2017-05-02
maxAge option incorrectly handles values expressed in number of milliseconds


zaak:
As stated in the docs:

> `maxAge`: the maximum allowed age for tokens to still be valid. Currently it is expressed in **milliseconds** or a string describing a time span zeit/ms. Eg: `1000`, `"2 days"`, `"10h"`, `"7d"`.  (...)

But if you pass a number, it's [converted to string](https://www.npmjs.com/package/ms#convert-from-milliseconds) by `ms` call [here](https://github.com/auth0/node-jsonwebtoken/blob/67550492cb3cff5fb764322563bb5864041eee5f/verify.js#L168), as there's no `typeof` check for `options.maxAge`. Because of this all the following validation logic is broken in this case.

ziluvatar:
Thanks @zaak!! There was an [issue](https://github.com/auth0/node-jsonwebtoken/issues/273) about how using milliseconds does not match quite well with JWT, however, we didn't realize that it actually does not work. The good part is that we can fix both at same time, making the code work with number (only for seconds) right away.

ziluvatar:
Fix: https://github.com/auth0/node-jsonwebtoken/pull/349

PR #349, opened 2017-05-13
maxAge: Fix logic with number + use seconds instead of ms

ziluvatar:
Fixes [#346](https://github.com/auth0/node-jsonwebtoken/issues/346)
Fixes [#273](https://github.com/auth0/node-jsonwebtoken/issues/273).

Since this will expect seconds to be used (even using string type), I mark it for `next-major`.

### DIFF
--- a/test/verify.tests.js
+++ b/test/verify.tests.js
@@ -68,8 +68,8 @@ describe('verify', function() {
   });
 
   describe('expiration', function () {
-    // { foo: 'bar', iat: 1437018582, exp: 1437018583 }
-    var token = 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJmb28iOiJiYXIiLCJpYXQiOjE0MzcwMTg1ODIsImV4cCI6MTQzNzAxODU4M30.NmMv7sXjM1dW0eALNXud8LoXknZ0mH14GtnFclwJv0s';
+    // { foo: 'bar', iat: 1437018582, exp: 1437018592 }
+    var token = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJmb28iOiJiYXIiLCJpYXQiOjE0MzcwMTg1ODIsImV4cCI6MTQzNzAxODU5Mn0.3aR3vocmgRpG05rsI9MpR6z2T_BGtMQaPq2YR6QaroU';
     var key = 'key';
 
     var clock;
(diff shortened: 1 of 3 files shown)


verify.js
```diff
@@ -165,15 +165,13 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
     if (typeof payload.iat !== 'number') {
       return done(new JsonWebTokenError('iat required when maxAge is specified'));
     }
-    // We have to compare against either options.clockTimestamp or the currentDate _with_ millis
-    // to not change behaviour (version 7.2.1). Should be resolve somehow for next major.
-    var nowOrClockTimestamp = ((options.clockTimestamp || 0) * 1000) || Date.now();
-    if (nowOrClockTimestamp - (payload.iat * 1000) > maxAge + (options.clockTolerance || 0) * 1000) {
-      return done(new TokenExpiredError('maxAge exceeded', new Date(payload.iat * 1000 + maxAge)));
+
+    var maxAgeTimestamp = timespan(options.maxAge, payload.iat);
```
mborst: Are you aware that your if `options.maxAge` was created by a `new String(...)` call, this will always return an error? Your `timespan` method does a `typeof x === 'string'` but `typeof new String('foo') === object`.

ziluvatar: I didn't think about that case, I reused one function we already have. I'll take a look to it.

mborst: I just stumbled over it when looking at the implementation of that function. I guess different rules apply now since it has to deal with external input.

ziluvatar: I just took the same approach used in `sign()` https://github.com/auth0/node-jsonwebtoken/blob/master/sign.js#L114-L116 so that we can cover when the format of the string itself is not correct.

mborst: I think it would be nicer for the user to generally test for string with something like
https://github.com/lodash/lodash/blob/master/isString.js, but that might better be done in a different PR.

ziluvatar: In this case `ms(new String('1s'))` gives you `Error: val is not a non-empty string or a valid number. val="1s"`. For other cases where it does not matter one or the other it makes totally sense. Since we are entering lodash type checks in other PR, we could check&modify the ones for strings with `isString`.

mborst: Ah, ok, so your way of doing it is 👍

dinvlad: Has it been/planned to be merged?

dinvlad: Also, why not simply ms(ms(options.maxAge))?

ziluvatar: I wanted to merge a bunch of "breaking changes" and then increase major version. It seems creators of the others PRs forgot about them (we asked for changes).
So, I'll try to find time to fix them probably, but I don't have ETA for now.

> why not simply ms(ms(options.maxAge))?

ms(ms(number)) => number
ms(ms(string)) => string
We want end up having the number.

dinvlad: Ok thanks, in that case ms(ms(ms())) might also do the trick...

dinvlad: Ok, thanks, got it.

dinvlad: To add to the conversation above, both `ms(String('1s'))` and `ms(String(1000))` work. No need to use new, as [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) object can be used as a global constructor for strings.

ziluvatar: Well, it is the consumer who is deciding what to send. But thanks to your comment I realized I didn't add a test for `String('12s')`, I'll add it later.

dinvlad: Makes sense - thanks for noting that!

test/verify.tests.js
```diff
@@ -114,70 +114,97 @@ describe('verify', function() {
...
+            assert.isUndefined(p);
+            done();
+          });
         });
       });
 
-      it('should not error for claims issued before a certain timespan but still inside clockTolerance timespan', function (done) {
-        clock = sinon.useFakeTimers(1437018582500);
-        var options = {algorithms: ['HS256'], maxAge: '321ms', clockTolerance: 100};
+      [String('6s'), '5s', 5].forEach(function (maxAge) {
```
fiddur: All the other multi-tests have different representations of the same time; this is different.  Seems like a mistake?

ziluvatar: Yep, copy&paste mistake. Fixing.

PR #352, opened 2017-05-17
Bump ms to v2 due a ReDoS vuln

dominykas:
ms@0.7.3 is vulnerable to a ReDoS attack: https://snyk.io/vuln/npm:ms:20170412

Looking at the code, the breaking change in ms@2.x that the accepted string is now limited to 100 chars.


ziluvatar:
Thanks @dominykas!

Issue #97, opened 2015-06-15
examples & tests for ES256


shea256:
Hi, are there any examples and/or tests for ES256?

I'm trying to sign a message with jwt.sign() and I would like to be able to pass in a plain hex private key, but in the example with RS256, a PEM is required.

Could you please advise? Thanks!

omsmith:
ES* also requires a PEM key (like RS*, it is calling into `crypto`/openssl).

ES* interoperability is currently broken thoug - see https://github.com/brianloveswords/node-jwa/issues/3 and https://github.com/brianloveswords/node-jws/issues/20

omsmith:
The interop issue I mention is fixed at this point, btdubs

shea256:
OK great thanks @omsmith. Recent commit / PR? And will the example be identical to RS256?

omsmith:
https://github.com/brianloveswords/node-jwa/pull/5 released in _node-jwa@1.0.1_ which is a indirect dependency of _jsonwebtoken_.

It works identically to RS256 yep, just pass in a PEM for EC instead. secp256r1 for ES256, secp384r1 for ES384, and secp521r1 for ES512.

shea256:
Oh wait, secp256k1 is not supported?

omsmith:
The [JWA spec](https://tools.ietf.org/html/rfc7518#section-3.4) defines _ES256_ as _ECDSA using P-256 and SHA-256_ where _P-256_ is another name for _secp256r1_

shea256:
Oh that's a pain. I actually think I've been using JWT libraries with SECP256K1. Not sure if there's something wrong there or they are actually going to both work. I'll experiment with it.

omsmith:
_jsonwebtoken_ and it's dependencies don't decode the keys and verify the correct one is being used, which means you could use a different curve, but it would be incorrect to call it "ES256" at that point. There's also the trouble that other libraries may not be so lenient, and this set of libraries may not be so lenient in the future, which means you're risking interop issues.

The set of specs that _JWT_ uses allows you to define your own "alg" for usage in your applications, but I don't know of any libraries across the different languages which support that case very well, which means you're on your own at that point.

ChristopherA:
Any progress on the question of a new alg=???? for keys using bitcoin's secp256k1?

gillesdemey:
The following works fine for me:

```
openssl ecparam -genkey -name secp256k1 -noout -out private.pem
[...]

PR #360, opened 2017-06-11
Add tests for ES256

ziluvatar:
It reuses same tests we have for RS256, for ES256 in this case.
It only wraps the main `describe` to run it once per asymmetric algorithm and replace `'RS256'` by `algorithm` variable. I didn't add/remove any test internally.

I uploaded a x509 certificate as well, it is not used, but it can be useful in the future.

Closes https://github.com/auth0/node-jsonwebtoken/issues/97


test/jwt.asymmetric_signing.tests.js
```diff
@@ -0,0 +1,449 @@
...
+
+        it('should be valid expiration', function (done) {
+          jwt.verify(token, pub, function (err, decoded) {
+            assert.isNotNull(decoded);
+            assert.isNull(err);
+            done();
+          });
+        });
+
+        it('should be invalid', function (done) {
```
fiddur: In the same describe `when signing a token with expiration` there is quite contradictory cases like `should be valid` and `should be invalid`.  Also, the describe creates a token, which is used in some cases and overwritten in other cases.  Either, make separate describe blocks for `when expiration has gone out` etc, or make the case-description clearer.  And decide if the `token` variable is for the whole describe or make one per case :)

ziluvatar: TBH I didn't check the tests internally, I just wanted to run the same ones for ES256 as well.
If you feel I should refactor the tests before applying these changes let me know, I can prepare a PR to refactor them + renaming and a following one with the wrapping for ES256.

PR #361, opened 2017-06-14
document keyid as option

rjruizes:
The keyid (`kid`) can be passed as an option to `sign()`, but it's not documented. 
https://github.com/auth0/node-jsonwebtoken/blob/master/sign.js#L59


ziluvatar:
Thanks @rjruizes catching that

PR #371, opened 2017-07-14
docs: about refreshing tokens

ziluvatar:
Before closing the issue and PR about this topic, we will add a note to the README and an example about how this can be done by consumers of the library.


fiddur:
Fixes #172.
Fixes #122.

Issue #373, opened 2017-07-24
Undefined private keys


anhldbk:
Hi there,

If I `sign` with an undefined key (when a callback is provided), I expect to have `err` to be a valid value. But currently, it's not. Here's the code:

```js
// test.js
const jwt = require("jsonwebtoken");

const doc = {
	id: 1
};
const key = undefined;

jwt.sign(doc, key, (err, token) => {
	console.log("-------------------------");
	console.log(err);
	console.log(token);
});
```

And the blank output:

```
Projects/play/test is 📦 v0.1.0 via ⬢ v7.10.0
➜ node test.js

Projects/play/test is 📦 v0.1.0 via ⬢ v7.10.0
➜ 
```

So there's something wrong, right?

Issue #286, opened 2016-12-22
Signing with undefined secret never gets resolved


robertrossmann:
Currently there is a [bug in node-jws](https://github.com/brianloveswords/node-jws/issues/62) where supplying `undefined` or other falsy value to `secret` results in the signing process never running to completion.

The solution for the problem is known and should probably be fixed upstream, but since this directly affects this library as well and given its popularity I thought I'd let you know. Perhaps you'd like to make sure `secret` has a value before handing it over to JWS for signing, I do not see much use of using `undefined` as my secret, anyway.

Good luck and merry Christmas! 🎁 🎅

ziluvatar:
After a breaking change detected on the fix developed for this issue, this bug remains open only for the case where the algorithm is `none` and the consumer is using callback-ish code style.
We'll need to think how to fix that or try to fix the bug in jws.

PR #374, opened 2017-07-24
sign: add check to be sure secret has a value

ziluvatar:
Lately there are more people raising this, not big deal but it can be weird and create confusion. We fix it in our layer, since the [bug in JWS is still open](https://github.com/brianloveswords/node-jws/issues/62).

In case of sync call, with current master code this error is returned: `TypeError: secret must be a string or buffer`.
With this PR both will return the same error `Error: secretOrPrivateKey must have a value`, I don't think the change would break anyone, because if someone got that error their integration would not be working at all.

Fixes: https://github.com/auth0/node-jsonwebtoken/issues/286
Fixes: https://github.com/auth0/node-jsonwebtoken/issues/373

### DIFF
--- a/sign.js
+++ b/sign.js
@@ -66,6 +66,9 @@ module.exports = function (payload, secretOrPrivateKey, options, callback) {
     throw err;
   }
 
+  if (!secretOrPrivateKey) {
+    return failure(new Error('secretOrPrivateKey must have a value'));
+  }
 
   if (typeof payload === 'undefined') {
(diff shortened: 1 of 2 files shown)


rhysmccaig:
This breaks the use of the none algorithm.

nguilford:
@rhysmccaig  The workaround is to pass in a non-zero length string. Not ideal behavior, but it unblocks you when using `none` algo.

PR #376, opened 2017-08-01
Update readme

xeodou:


ziluvatar:
Sorry @xeodou I don't fully get why that change is needed, can you explain the reason?
`RS256` is `RSA SHA256`, saying `RSA RS256` is redundant, isn't it?

xeodou:
yes, @ziluvatar It's my mistake. Thanks for your reply, I'll close the PR.

PR #378, opened 2017-08-04
Added algorithm aliases to support Microsoft's JwtSecurityTokenHandler.

bmoelk:
Reference: https://msdn.microsoft.com/en-us/library/system.identitymodel.tokens.jwtsecuritytokenhandler.inboundalgorithmmap(v=vs.114).aspx


bmoelk:
Sorry, I don't think this should be pulled. :)

PR #380, opened 2017-08-11
README fixes

oughter:
- Fix typo in `options.header` reference
- Update `Buffer()` example because `new Buffer()` is [deprecated](https://nodejs.org/api/buffer.html#buffer_new_buffer_string_encoding)


ziluvatar:
Thanks @oughter ! I wasn't sure about the Buffer one since that's not available on node v4, anyway it's fine to give an example assuming latest version (avoiding deprecated usages)

Issue #381, opened 2017-08-11
Pull Request #374 has broken the use of the 'none' algorithm.


rhysmccaig:
Noticed this in my own test cases - when using the none algorithm the sign() function now throws an error if a secret isn't defined. (which is find for `alg: none`)

I believe the simple workaround which preserves the workaround added in #374 is to simply set the secretOrPrivateKey to any string when the algorithm is set to `none`. Will add a PR shortly.

PR #382, opened 2017-08-12
Fix for #381. Set secret string before using jws when alg is none

rhysmccaig:
Fix for https://github.com/auth0/node-jsonwebtoken/issues/381.
- When secret is Falsy, but algorithm is set to none, make secret a non zero length string before using jws.sign()/createSign()
- Added 2 unit tests

### DIFF
--- a/sign.js
+++ b/sign.js
@@ -67,7 +67,11 @@ module.exports = function (payload, secretOrPrivateKey, options, callback) {
   }
 
   if (!secretOrPrivateKey) {
-    return failure(new Error('secretOrPrivateKey must have a value'));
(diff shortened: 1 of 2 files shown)


sign.js
```diff
@@ -67,7 +67,11 @@ module.exports = function (payload, secretOrPrivateKey, options, callback) {
   }
 
   if (!secretOrPrivateKey) {
```
ziluvatar: What if we modify the if statement to take the that situation into account? Something like `(!secretOrPrivateKey && options.algorithm !== 'none')` what do you think?

PR #386, opened 2017-08-17
Fix breaking change on 7.4.2 for empty secret + "none" algorithm (sync code style)

ziluvatar:
Based on PR: https://github.com/auth0/node-jsonwebtoken/pull/382
Fix: https://github.com/auth0/node-jsonwebtoken/issues/381

It modifies @rhysmccaig approach to simply check the algorithm (`none`?) before raising an error. I think modifying the secret is hacky.

So this will reopen https://github.com/auth0/node-jsonwebtoken/issues/286 for the case of callback-ish code style and algorithm `none` (it didn't work before either). But it will solve the breaking change introduced by https://github.com/auth0/node-jsonwebtoken/pull/374

### DIFF
--- a/sign.js
+++ b/sign.js
@@ -67,7 +67,11 @@ module.exports = function (payload, secretOrPrivateKey, options, callback) {
   }
 
   if (!secretOrPrivateKey) {
-    return failure(new Error('secretOrPrivateKey must have a value'));
+    if (options.algorithm === 'none') {
+      secretOrPrivateKey = 'Fix for https://github.com/auth0/node-jsonwebtoken/issues/381';
+    } else {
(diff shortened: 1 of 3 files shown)


rhysmccaig:
Thanks for this - I had explored this approach, but for some reason I thought it still failed due to the way JWS/JWA worked, looks like I was wrong!

ziluvatar:
@rhysmccaig thank you for your fast PR!! (sorry for putting my suggestion on top of your code, but I worried you may not notice the comment and I didn't want to leave a breaking-change out there)

You were right, I reopen the [original issue](https://github.com/auth0/node-jsonwebtoken/issues/286#issuecomment-323097451), although now it has smaller scope. I think we need to fix jws to get rid of it.

PR #394, opened 2017-09-10
Remove `lodash.isarray` dependency

chmelevskij:
Npm gives deprecation warning. Also `Array.isArray` is already used in library.


ziluvatar:
@chmelevskij thanks for noticing and sending this PR right away 👏 
Released on: v8.0.1

PR #398, opened 2017-09-18
Enhance audience check to verify against regular expressions

TheBusCantSwim:
The content of this pull request is to introduce the possibility to verify the audience against a regular expression.

The usage of this feature would be in a restrictive scenario where the audience needs to be verifiable against a large set of audiences because it has a parameter as part of the urn.

E.g.: Using a rest-endpoint to query all items of a specific order:
The endpoint listens to uris of the following structure: /Orders/:OrderId/Items
With the old implementation the audience could be verified against:
- a subset of the uri: '/Order/Items'
- an exhaustive list of permuatations ['/Order/1/Items', '/Order/2/Items', ... '/Order/1234/Items']

Or with the new implementation it can be verified against valid orderIds: ^\/Order\/\d{1,4}\/Items$

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues
 `options`
 
 * `algorithms`: List of strings with the names of the allowed algorithms. For instance, `["HS256", "HS384"]`.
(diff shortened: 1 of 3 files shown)


ziluvatar:
It looks good to me, good improvement. Can you update the README with this change?

ziluvatar:
Released on v8.1.0. Really nice the feature you added 👏  @TheBusCantSwim

Issue #385, opened 2017-08-14
Little mistake in README


shawninder:
I found a little mistake in the README, see https://hyp.is/49FjfoEDEee2WVfxwy1Brw/www.npmjs.com/package/jsonwebtoken

> Callback is passed the decoded payload if the signature and optional expiration, audience, or issuer are valid.

I think there is a part missing here. I'd like to propose the following new sentence as a replacement:

> Callback is passed the decoded payload if the signature is valid and optional expiration, audience, or issuer are valid.

ziluvatar:
I understand almost the same (actually what I don't like is the "Callback is passed the decoded payload").
Anyway, I see it's clearer your way, be free to open a PR and I'll merge it.

joovel:
@ziluvatar : I've created the PR.

ziluvatar:
PR was merged.

PR #408, opened 2017-09-27
#385: Tweak README

joovel:
Fixes issue #385 , adds a few other definite articles for clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ $ npm install jsonwebtoken
 
 ### jwt.sign(payload, secretOrPrivateKey, [options, callback])
 
-(Asynchronous) If a callback is supplied, callback is called with the `err` or the JWT.
+(Asynchronous) If a callback is supplied, the callback is called with the `err` or the JWT.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,7 +106,7 @@ jwt.sign({
 
 ### jwt.verify(token, secretOrPublicKey, [options, callback])
 
-(Asynchronous) If a callback is supplied, function acts asynchronously. Callback is passed the decoded payload if the signature and optional expiration, audience, or issuer are valid. If not, it will be passed the error.
+(Asynchronous) If a callback is supplied, function acts asynchronously. The callback is passed the decoded payload if the signature is valid and optional expiration, audience, or issuer are valid. If not, it will be passed the error.
```
ziluvatar: Good change, however, this sentence always feels weird to me, don't you think it should use the same "pattern" as we use for `sign`? 
**The callback is called with the** decoded payload... . If not, it will be **called with** the error.

joovel: Updated. I also made the wording between the Synchronous and Asynchronous section the same.

Issue #403, opened 2017-09-26
version new jsonwebtoken error


thandonguocmo2020:
I use the new version with the error occurred  "jsonwebtoken": "^8.0.1",.

I use the old version the bug has been removed   "jsonwebtoken": "^7.4.3".

MESSAGE ERROR ==> 

``` code
      var token;
                        try {
                            token = jwt.sign(doc, jwtSecret, { expiresIn: '10h' });
                        } catch (error) {
                          console.log("ERROR==================>"+ JSON.stringify(error));     
                        }
           

```

events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: Expected object
    at validate (E:\App\code\node_modules\jsonwebtoken\sign.js:34:11)
    at Object.module.exports [as sign] (E:\App\code\node_modules\jsonwebtoken\sign.js:100:7)
    at E:\App\code\dist\app.js:402:37
    at Query.<anonymous> (E:\App\code\node_modules\mongoose\lib\model.js:3841:16)
    at E:\App\code\node_modules\kareem\index.js:273:21
    at E:\App\code\node_modules\kareem\index.js:131:16
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickDomainCallback (internal/process/next_tick.js:128:9)

ziluvatar:
We have changed the validators used for the `sign` method. Take a look to the first item in the migration guide: https://github.com/auth0/node-jsonwebtoken/wiki/Migration-Notes:-v7-to-v8

We use lodash now, take a look to this function: https://lodash.com/docs/4.17.4#isPlainObject

Anyway I think we should communicate better that error, it could say something like "expected payload object to be a plain object" or similar.

Can you share the content of `doc`?

joovel:
@ziluvatar : I've added a PR to change the wording.

ziluvatar:
PR was merged. It's on v8.1.0

UNADYNE:
Can this be fixed by simply adding `.toJSON()`?

kedrovski:
You can try it `jwt.sign(doc.**toObject()**, jwtSecret);`

NodeJoSe:
@kedrovski solution is fine but i suggest you to use: `JSON.parse(JSON.stringify(doc))` if your doc object comes from diferent sources, for example:
 
You could have to pass diferent "doc" objects, one returned from mongoose and another defined by you like this:

`doc = { 
    name: 'foo', 
    description: 'bar' 
[...]

PR #409, opened 2017-09-27
#403: Clarify error wording.

joovel:


sign.js
```diff
@@ -31,7 +31,7 @@ var registered_claims_schema = {
 
 function validate(schema, unknown, object) {
   if (!isPlainObject(object)) {
-    throw new Error('Expected object');
+    throw new Error('Expected payload to be a plain object.');
```
ziluvatar: This can be raised from payload or options validation.

joovel: I've made some changes to distinguish these.

ziluvatar:
Released on v8.1.0. 👏  Thanks again @joovel !!

Issue #402, opened 2017-09-22
Using Firefox and Safari, jwt.verify raises Error.capturestacktrace is not a function


TLX3:
I've started testing my application outside of Chrome and started receiving the error regarding Error.capturestacktrace is not a function.

I have both the latest Firefox and Safari. The token verification works fine on the latest version of Chrome. This error occurs specifically after the token is expired and I attempt to verify it. 

The token verification below is executed during specific route transitions on the client side. The  error caused prevents me from logging inside of the verification callback.

```javascript
 jwt.verify(
      token,
      secret,
      function(err, verifiedJwt) {
        if (err) {
          if (
            err.name === "TokenExpiredError" ||
            err.name === "JsonWebTokenError"
            ) {
            next("/logout");
        }
      } else {
          //Refresh token if it is not expired
          store.dispatch("REFRESH_TOKEN", token);
        }
      }
  );
```

joovel:
@ziluvatar : I've added a PR to ensure that `captureStackTrace` is truthy before attempting to call it.

joovel:
@ziluvatar : I'm also curious if browser support is a thing for this project ... both this repo and jws specifically call out node, but there's no explicit statement in the README whether browsers are supported or not.

ziluvatar:
This project focus is NodeJS, nevertheless since the community seems to be using it for browsers, we balance the issue, the effort to maintain node vs browser and it makes sense.

ziluvatar:
PR was merged. It's on v8.1.0

TLX3:
Thanks!

PR #410, opened 2017-09-27
#402: Don't fail if captureStackTrace is not a function

joovel:


lib/JsonWebTokenError.js
```diff
@@ -1,6 +1,8 @@
 var JsonWebTokenError = function (message, error) {
   Error.call(this, message);
-  Error.captureStackTrace(this, this.constructor);
+  if(Error.captureStackTrace) {
+    Error.captureStackTrace(this, this.constructor);
+  }
```
ziluvatar: I'm curious, have you investigated why this happens in Chrome? (it uses the same v8 as node). Could you take a look? Maybe we can improve the Error creation without this `if`.

joovel: From the issue:

> I've started testing my application outside of Chrome and started receiving the error

So my understanding is that this function is specific to V8, meaning it _works_ Chrome and node and _not_ FF and Safari.

Here are some options: 
* http://codegists.com/snippet/javascript/errorsjs_girvo_javascript for a slightly better in-code solution
* https://www.npmjs.com/package/error-polyfill for a heavier-weight browser-only polyfill

ziluvatar: :( I misread it, it's simpler your approach. I'll approve, I'll wait for merging to see if we have more fixes before I release a patch version (I mean currently).

ziluvatar:
Released on v8.1.0. 🎉  Thanks @joovel !!

PR #428, opened 2017-12-05
add newer node versions to build matrix

DanielMSchmidt:
Hi there,

Thanks for the great project 👍  
We rely on newer node versions currently, so I thought adding this might help ensure that no breaking changes are introduced by accident.


ziluvatar:
Thanks @DanielMSchmidt !

Issue #427, opened 2017-11-29
Adding custom algorithm


Teukros:
I want to use this package in the new project, but in some cases we use our own custom algorithm. Would you accept adding some logic that allows to register new algorithm?

ziluvatar:
Do you mean your own signing / verification algorithm?

ziluvatar:
Sorry I just saw your PR https://github.com/auth0/node-jsonwebtoken/pull/429
I'm not sure about it, in case of adding it, it should not be 'custom' algorithm. As it is mentioned in JWA it should be a qualified name to avoid collisions https://tools.ietf.org/html/rfc7515#page-10.

Maybe using something similar to how passport or oauth2orize library work. Where you could do like (just wondering):
`jwt.useAlg('http://mysite.com/blah', signFunction, verifyFunction)`

Curiosity, what is the reason behind implementing your own algorithm?

Teukros:
Of course it can be renamed from "custom" to whatever you prefer. I could also add another option like boolean "customAlgorithm" instead of passing algorithm name. Another solution is also to register a new algorithm right on the beginning when you require the library, but it would need bigger intervention in code and storing global (in scope of lib) object with algorithms used so I don't like it.

The reason why I want to implement my own algorithm is to use the library in every case when I handle request. Sometimes requests need HS256, RS256 encoding, and sometimes it will be our own algorithm (let's name it ABC). Thanks to this change, I can use jsonwebtoken lib in every case, using the same function. If it's ABC alg, I will have to pass additional option with algorithm function, but it's still the same logic.

I tried to find your example with jwt.useAlg in passport & oauth2orize libs, I wasn't able to find them - could you please tell me where is it?

rochdev:
Not sure if this is the right issue for this, but we have a very similar use case where we want to use an existing algorithm (RS256) but do the signing through an external party (Google Cloud) where we never have the private key.

With the syntax @ziluvatar have mentioned it would work in our case by doing something like this:

```js
jwt.useAlg('RS256', () => iam.projects.serviceAccounts.signBlob(...))
```

Since only Google can verify the signature in this case the verifyFunction should probably be optional.

ziluvatar:
I think your case @rochdev is slightly different, 'RS256' is supported by this library, we would need to analyze better that approach.

@Teukros I meant this library https://github.com/jaredhanson/passport
There you have separation between configuration and usage (`passport.use()` vs `passport.authenticate()`)

So you could have once:
```
[...]

PR #429, opened 2017-12-06
Adding custom algorithm

Teukros:
This pull request introduce possibility to use your own algorithm with jsonwebtoken library. If you pass proper options you can use sign, verify and decode methods together with chosen algorithm function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ encoded private key for RSA and ECDSA. In case of a private key with passphrase 
 * `algorithm` (default: `HS256`)
 * `expiresIn`: expressed in seconds or a string describing a time span [zeit/ms](https://github.com/zeit/ms). Eg: `60`, `"2 days"`, `"10h"`, `"7d"`
 * `notBefore`: expressed in seconds or a string describing a time span [zeit/ms](https://github.com/zeit/ms). Eg: `60`, `"2 days"`, `"10h"`, `"7d"`
+* `customAlgorithmFunction`: function which will be used if custom algorithm is chosen
 * `audience`
 * `issuer`
 * `jwtid`
@@ -71,6 +72,25 @@ var token = jwt.sign({ foo: 'bar' }, cert, { algorithm: 'RS256'});
 jwt.sign({ foo: 'bar' }, cert, { algorithm: 'RS256' }, function(err, token) {
   console.log(token);
 });
+
+// sign asynchronously and synchronously with custom algorithm
+function customAlgorithmFunctionAsync(payload, secretOrPrivateKey, options, callback) {
+    return callback(null, result);
+}
+var optAsync = {
+    algorithm: 'custom',
+    customAlgorithmFunction: customAlgorithmFunctionAsync
+};
+jwt.sign({ foo: 'bar' }, 'secret', optAsync, callback);
+
+function customAlgorithmFunctionSync(payload, secretOrPrivateKey, options) {
+    return result;
+}
+var optSync = {
+    algorithm: 'custom',
+    customAlgorithmFunction: customAlgorithmFunctionSync
+};
+jwt.sign({ foo: 'bar' }, "secret", optSync);
 ```
 
 #### Token Expiration (exp claim)
@@ -128,6 +148,7 @@ As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues
 * `clockTolerance`: number of seconds to tolerate when checking the `nbf` and `exp` claims, to deal with small clock differences among different servers
 * `maxAge`: the maximum allowed age for tokens to still be valid. It is expressed in seconds or a string describing a time span [zeit/ms](https://github.com/zeit/ms). Eg: `1000`, `"2 days"`, `"10h"`, `"7d"`.
 * `clockTimestamp`: the time in seconds that should be used as the current time for all necessary comparisons.
+* `customAlgorithmFunction`: function which will be used to verify signature if algorithm is custom.
 
 
 ```js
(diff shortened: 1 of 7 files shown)


sign.js
```diff
@@ -13,14 +14,15 @@ var sign_options_schema = {
...
+  algorithm: { isValid: includes.bind(null, ['RS256', 'RS384', 'RS512', 'ES256', 'ES384', 'ES512', 'HS256', 'HS384', 'HS512', 'custom', 'none']), message: '"algorithm" must be a valid string enum value' },
   header: { isValid: isPlainObject, message: '"header" must be an object' },
   encoding: { isValid: isString, message: '"encoding" must be a string' },
   issuer: { isValid: isString, message: '"issuer" must be a string' },
   subject: { isValid: isString, message: '"subject" must be a string' },
   jwtid: { isValid: isString, message: '"jwtid" must be a string' },
   noTimestamp: { isValid: isBoolean, message: '"noTimestamp" must be a boolean' },
-  keyid: { isValid: isString, message: '"keyid" must be a string' }
+  keyid: { isValid: isString, message: '"keyid" must be a string' },
+  customAlgorithmFunction: {isValid: isFunction, message: "custom algorithm must provide it's own sign function"}
```
montgomery1944: it's -> its
" -> '

sign.js
```diff
@@ -169,20 +171,30 @@ module.exports = function (payload, secretOrPrivateKey, options, callback) {
...
-    jws.createSign({
-      header: header,
-      privateKey: secretOrPrivateKey,
-      payload: payload,
-      encoding: encoding
-    }).once('error', callback)
-      .once('done', function (signature) {
-        callback(null, signature);
+    if (isCustomAlgorithmUsed) {
+      options.customAlgorithmFunction(payload, secretOrPrivateKey, options, function(err, result) {
```
montgomery1944: Formatting.

verify.js
```diff
@@ -29,6 +30,15 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
     };
   }
 
+  if (options.algorithms && includes(options.algorithms, 'custom')) {
+    try {
+      decodedToken = options.customAlgorithmFunction(jwtString, secretOrPublicKey, options);
+      return done(null, decodedToken);
+    } catch(err) {
```
montgomery1944: Formatting.

README.md
```diff
@@ -203,6 +234,8 @@ __Warning:__ This will __not__ verify whether the signature is valid. You should
 
 * `json`: force JSON.parse on the payload even if the header doesn't contain `"typ":"JWT"`.
 * `complete`: return an object with the decoded payload and header.
+* `isCustomAlgorithmUsed`: boolean that declares if you want to use your own algorithm for decoding.
+* `customAlgorithmFunction`: function which will be used for decoding, if isCustomAlgorithmUsed is true
```
montgomery1944: Missing ".".

rochdev:
I think it should be possible to use a custom signing function even when using a predefined algorithm.

My use case: https://github.com/auth0/node-jsonwebtoken/issues/427#issuecomment-356363954

ziluvatar:
Sorry, I still have some concerns about this, what is the benefit of adding custom signing to this library? (at least how it is in the PR so far)

- On signing: I can see a little: some payload generation. Signer ends up even filling the header part. The function could even avoid adding any header or anything it could bypass the JWT format and build any other format.
- On verifying: It just get a string and return the object.

It seems to me like a simple serializer / deserializer. We should think how to offer this as simple as possible to the consumer and be sure that we are still in JWT playground.

If you take a look to `jws` package, what you want is to use your own `algo`: https://github.com/brianloveswords/node-jws/blob/master/lib/sign-stream.js#L21, so to generate the last part of the JWT (header.payload.**signature**). Is that right?

There is an open ticket there about your own use case: https://github.com/brianloveswords/node-jws/issues/34. I think it would have more sense to have it supported in there.

If we still wanted to support it here we would need to write, or bring, most of that to this library, having here two signers with same interface `custom` and `jws` and leaving the final custom signing function for the consumer similar to what they do in `jws`: `algo(securedInput, secretOrKey)` where `securedInput` is "header.payload" in base64 with encoding in mind.
If we don't do all of that, consumer would need to build `jws` basically to do it right.

What if you try to bump up the topic on jws repo and we pause this for now? Thoughts?

novemberborn:
> what is the benefit of adding custom signing to this library?

I'm currently investigating custom algorithms. In my use case I have a secp256k1 private key I want to use to sign a JWT. The JWT would contain a claim of the corresponding public key.

Interface wise I think the `algorithms` option here should be an array that accepts strings and objects. For objects, keys would be the custom algorithm name and the value either a signer or verifier:

```js
jwt.sign(payload, secretOrPrivateKey, {
  algorithm: {
    'urn:foo:bar:custom:yay' (input) {
      return sign(input)
    }
  }
})

jwt.verify(token, publicKey, {
  algorithms: [
    {
      'urn:foo:bar:custom:yay' (input) {
        return verify(input)
      }
    }
  ]
})
```

This library could select the correct `algorithm` when verifying. I agree that it makes sense for `jws` to accept a custom function. Though that project doesn't look particularly active.

ziluvatar:
There was a PR about it in the past: https://github.com/brianloveswords/node-jws/pull/35 I bet it was closed since some feedback was provided but no reaction was made by the PR creator.
Maybe someone can re-create it applying the feedback.

panva:
JWT algorithms are well defined and we're committed to support the defined ones, as soon as the `jws` dependency supports them too.

`secp256k1` curve EC keys and alg name registration is only just being [standardized](https://tools.ietf.org/html/draft-ietf-cose-webauthn-algorithms-00) and is still very much in [flux](https://mailarchive.ietf.org/arch/msg/cose/91MouVA43DefbpQOB7l5daCEeIc)

The only other missing defined DSA JWA is EdDSA which is available in node 12 and will eventually make it to `jsonwebtoken` too.

Also as a general recommendation you should not roll your own crypto, which this very much looks like.

Issue #435, opened 2017-12-21
Signing a payload with a `iat` results in an invalid `nbf` value


MitMaro:
Related Issue: #142 
***
## Problem Description

The calculation of `payload.nbf` does not use `payload.iat` when `notBefore` is passed an offset, and instead the current time is always used.

## Expected

In this case I would expect that the `payload.nbf` value to be an offset based on the passed `payload.iat`. This would be consistent with `payload.exp`.

## Reduced Test Case

```
const {sign, decode} = require('jsonwebtoken');
const token = sign({foo: 'bar', iat: 100}, 'secret', {notBefore: '-1s'});
const decoded = decode(token);
console.log(`Expecting nbf value ${decoded.nbf} to equal 99`);
> Expecting nbf value 1513884708 to equal 99
```
## Investigation

It seems that the `timestamp` is not passed to `timespan` function on [Line 148 of sign.js](https://github.com/auth0/node-jsonwebtoken/blob/bb27eb346f0ff675a320b2de16b391a7cfeadc58/sign.js#L148) similar to how it is passed to the call on [Line 155 of sign.js](https://github.com/auth0/node-jsonwebtoken/blob/bb27eb346f0ff675a320b2de16b391a7cfeadc58/sign.js#L155).

MitMaro:
I would be willing to put up a PR fixing the issue with tests if this is not by designed.

ziluvatar:
@MitMaro I think this simply wasn't fixed when the timestamp was introduced for `exp`. https://github.com/auth0/node-jsonwebtoken/issues/217. Be free to send a PR.

PR #437, opened 2018-01-11
Not Before (nbf) calculated based on iat/timestamp

MitMaro:
See #435 

Set nbf value based on the timestamp.

### DIFF
--- a/test/iat.tests.js
+++ b/test/iat.tests.js
@@ -12,4 +12,13 @@ describe('iat', function () {
     expect(result.exp).to.be.closeTo(iat + expiresIn, 0.2);
   });
 
+  it('should work with a nbf calculated based on numeric iat', function () {
+    var dateNow = Math.floor(Date.now() / 1000);
+    var iat = dateNow - 30;
+    var notBefore = -50;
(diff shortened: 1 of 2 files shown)


test/iat.tests.js
```diff
@@ -12,4 +12,13 @@ describe('iat', function () {
...
 
-});
\ No newline at end of file
+  it('should work with a nbf calculated based on numeric iat', function () {
+    var dateNow = Math.floor(Date.now() / 1000);
+    var iat = dateNow - 30;
+    var notBefore = -50;
+    var token = jwt.sign({foo: 123, iat: iat}, '123', {notBefore: notBefore});
+    var result = jwt.verify(token, '123');
+    expect(result.nbf).to.be.closeTo(iat + notBefore, 0.2);
```
ziluvatar: since the `timestamp` will be the `iat`, shouldn't it be equal always (instead of "closeTo")

MitMaro: I agree completely! I was following the previous test example to keep the style the same, but I will gladly update.

MitMaro: Changed to check for direct equals.

PR #440, opened 2018-01-24
Add payload field to defined errors

mtpc:
In some case we still want to access the parsed payload despite the token being considered invalid for any reason whatsoever.

### DIFF
--- a/lib/JsonWebTokenError.js
+++ b/lib/JsonWebTokenError.js
@@ -1,10 +1,11 @@
-var JsonWebTokenError = function (message, error) {
+var JsonWebTokenError = function (message, error, payload) {
   Error.call(this, message);
   if(Error.captureStackTrace) {
     Error.captureStackTrace(this, this.constructor);
   }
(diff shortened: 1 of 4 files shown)


ziluvatar:
Can you bring some example? Because you should not trust any information of the payload if it was considered invalid.

mtpc:
For example the token might have expired, but we still want to access its payload for some token renewal logic.
It might indeed not be a good idea to trust the information of an invalid token, but that's ultimately a decision to be taken by the application logic.

ziluvatar:
Can't you simply call `jwt.decode` in case of error? That would solve it without making the lib expose a payload that the lib consider invalid/no-trusted.

mtpc:
Generally speaking, the whole problem is to access the parsed payload in case of error without having to decode the token *once again*. A better solution than exposing it in errors might exist though, you're right.

ziluvatar:
Since the normal execution is not to get an error, I think if someone is interested in some information in that case it wouldn't be a pain to decode it.

PR #444, opened 2018-02-13
Check payload is not null when decoded.

### LINKED ISSUE #430: Check payload before trying to get nbf
In verify.js, I think there is an issue with the payload decoding. Or at least with robustness.
I receive some errors (`TypeError: Cannot read property 'nbf' of null`) with the following lines (Around line 110 of verify.js):

``
  var payload;

  try {
    payload = decode(jwtString);
  } catch(err) {
    return done(err);
  }

  if (typeof payload.nbf !== 'undefined' && !options.ignoreNotBefore) {
//...
``

The function decode sometimes (if the jwtString is malformed ?) returns null, isn't a good idea to check the payload before trying to use the nbf ?
It's a simple if in the try-catch, I can do a PR if needed.

Regards,

Gp2mv3

Gp2mv3:
Fixed "Cannot read property 'nbf' of null" fixes #430

### DIFF
--- a/decode.js
+++ b/decode.js
@@ -10,7 +10,7 @@ module.exports = function (jwt, options) {
   if(typeof payload === 'string') {
     try {
       var obj = JSON.parse(payload);
-      if(typeof obj === 'object') {
+      if(obj && typeof obj === 'object') {
         payload = obj;
       }
     } catch (e) { }


decode.js
```diff
@@ -10,7 +10,7 @@ module.exports = function (jwt, options) {
   if(typeof payload === 'string') {
     try {
       var obj = JSON.parse(payload);
-      if(typeof obj === 'object') {
+      if(obj && typeof obj === 'object') {
```
ziluvatar: Can you add a test?

MitMaro: Would it make more sense to check that `obj !== null` here instead? Might make it more clear that the additional condition is to guard against `typeof null => 'object'`.

ziluvatar:
@Gp2mv3 this was released on v8.2.1 🎉 

Thanks!!!

Issue #357, opened 2017-05-29
Expose function to apply options onto a payload


jondubois:
Right now the `sign` function applies the options to the payload internally but there is no way to get back that payload (with all the exp, iss, ... claim properties) without first encoding and then decoding the token (which is asynchronous and uses extra CPU).

It would be nice if either the sign function could somehow return the transformed (unsigned) payload after the options have been applied to it (but before encoding) or at least expose the apply-options logic as a function so that I can know what the transformed payload looks like synchronously without having to decode it as a separate async step.

The use case is that I need to have a reference to the JWT payload (with all its claims) as soon as the JWT is created (synchronously).

Right now the only way to get the transformed JWT payload synchronously is to duplicate the options mapping logic from jsonwebtoken in my own code.

jondubois:
A simple solution would be to add an option (e.g. passed to the sign method) to not clone/xtend the payload here: https://github.com/auth0/node-jsonwebtoken/blob/e202c4fd00c35a24e9ab606eab89186ade13d0cc/sign.js#L79 and instead to allow the payload argument to be mutated.

ziluvatar:
🤔 I think it'd be better, as you said in the first comment, to expose i.e: `generatePayload()` public function to get the jwt claims and make the `sign()` function use it internally.

jondubois:
@ziluvatar Do you know if there has been any progress with that since version 8?

I was thinking that it would be nice if there was also a publicly exposed `signPreparedPayload()` function which was used internally by the `sign()` function but unlike the `sign` function, it would expect a prepared payload (with all claims already prepared) as argument.

The internal `generatePayload()` would end up getting called twice in some situations which is wasteful.

ziluvatar:
No, no progress so far :( . You are right, however, all of that will add more complexity, I didn't thought about the double call to `generatePayload()` in my previous comment. Then I guess it would be better, as you said before, to pass an option like `mutatePayload` or similar, which by default would behave like nowadays (avoiding mutation).

If you prepare the PR I'd be happy to merge it (add docs and tests too).

PR #446, opened 2018-02-21
Add a new mutatePayload option

jondubois:
The `mutatePayload` option allows you to get an updated reference to the raw token payload after claims have been applied to it but before it has been encoded. Default behaviour is unchanged since `mutatePayload` defaults to false.

Added tests.

Closes https://github.com/auth0/node-jsonwebtoken/issues/357

Another simple (maybe more elegant) solution could have been to pass the signed-but-not-encoded payload as an extra argument to the callback but that wouldn't work with the synchronous version of the function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ encoded private key for RSA and ECDSA. In case of a private key with passphrase 
 * `noTimestamp`
 * `header`
 * `keyid`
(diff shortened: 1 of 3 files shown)


ziluvatar:
Thanks @jondubois , released on v8.2.0

Frondor:
How does this exactly works? What's the final example of implementing this with current `,sign(payload, secret)`? Would be nice to have an example in readme

PR #447, opened 2018-02-22
xtend stomps, cannot be used as a cloning function

ekristen:
Recently I discovered that jsonwebtoken was stomping all over the verify options. It turns out xtend is not safe to  use as a clone function. It does not create a new object that is safe for mutating.

This PR replaces xtend with lodash.merge and lodash.clone.

verify.js now uses clone to get a new object that doesn't stomp.
sign.js now uses merge instead of extend to merge options together.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,15 +20,16 @@
   },
   "dependencies": {
     "jws": "^3.1.4",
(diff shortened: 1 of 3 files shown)


ekristen:
Disregard, I was stomping locally.

PR #462, opened 2018-04-19
Replace isPlainObject with isObjectLike

kevincollins7:
This pull request replaces the use of lodash.isPlainObject with lodash.isObjectLike. There seems to be some  reported issues since the change to the object validation to use lodash. We were running into issues using node-postgres which returns an anonymous object as query results such as:
```
const user = await knex('user')
  .first('id', 'name', 'email')
  .where({ 'email: email });
```
Which would return something like this:
```
anonymous {
  id: '1234',
  name: 'John Doe'
}
```
Which failed the isPlainObject validation but passes the isObjectLike validation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "lodash.isinteger": "^4.0.4",
     "lodash.isnumber": "^3.0.3",
     "lodash.isobjectlike": "^4.0.0",
-    "lodash.isplainobject": "^4.0.6",
     "lodash.isstring": "^4.0.1",
     "lodash.once": "^4.0.0",
     "ms": "^2.1.1",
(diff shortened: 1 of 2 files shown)


ziluvatar:
Nice catch.
Taking a look to what `isObjectLike` allows and what `isPlainObject` or `joi.object()` allowed you should need to fail when an array is used.

Could you add tests for the different use cases? ie: 
* Valid: `{}`, `Object.create(null)`, `new (function Payload() { this.some = 'data' })()`, `{ some: 'data' }`
* Invalid: [1, 2]

And any other you can think of. That way if we change to another library in the future we won't break these scenarios again.

PR #468, opened 2018-05-18
Added Istanbul to check test-coverage

JacoKoster:
While looking to add tests for PR 413 i noticed it would be helpful to have an overview of the tested code. Hopefully this helps :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 node_modules
 .DS_Store
+.nyc_output
+coverage
+node_modules_bak
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@ node_modules
 .DS_Store
 .nyc_output
 coverage
-node_modules_bak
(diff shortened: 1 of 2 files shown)


JacoKoster:
Build has failed because nyc is not compatible with node < 5... :-(

ziluvatar:
I'll leave this on hold until we remove support for node 4 and 5. (I left a comment in your other PR https://github.com/auth0/node-jsonwebtoken/pull/470)

JacoKoster:
Because of the update of the CI, this can now be merged :-) 

(It is only a dev-dependency anyway)

.gitignore
```diff
@@ -1,2 +1,5 @@
 node_modules
-.DS_Store
\ No newline at end of file
+.DS_Store
+.nyc_output
+coverage
+node_modules_bak
```
MitMaro: Should this be here? Seems like it might have been included accidentally.

JacoKoster: Ah, yes, because the cost-of-modules takes quite some time i cancelled it and it left this folder. I added it in the ignore as i figured this shouldn't be checked in. I will remove it from the gitignore ;-)

PR #470, opened 2018-05-18
Updated travis config according to the latest release-schedule of node

JacoKoster:
Versions 4, 5 and 7 are now end of life.

See: https://github.com/nodejs/Release#release-schedule

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: node_js
 sudo: false
 before_install: npm i -g npm
 node_js:
+  - "10"
   - "8"
-  - "7"
   - "6"
-  - "5"
-  - "4"


ziluvatar:
@JacoKoster we can drop support on next major. I added the label to avoid forgetting.

JacoKoster:
Cool, what is the planning for the next major version?

ziluvatar:
Not real plan, whenever is needed, currently only this change (using istanbul for coverage reporting) is waiting for it. I'd like to wait in case we get more changes that need to break compatibility.

panva:
No need to stop testing with older versions until we actually decide to completely remove their support with the next major.

PR #480, opened 2018-06-03
Secret callback revisited

JacoKoster:
Based on [PR 413](https://github.com/auth0/node-jsonwebtoken/pull/413), without breaking existing functionality.

Added features

- verify's secretOrToken parameter can now be a callback function, see updated README.md for details
- updated tests and readme

Performance improvements
- merged two decodes into one

Dependencies
- Removed xtend and use Object.assign instead

See issue [406](https://github.com/auth0/node-jsonwebtoken/issues/406)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ $ npm install jsonwebtoken
 
 `secretOrPrivateKey` is a string, buffer, or object containing either the secret for HMAC algorithms or the PEM
 encoded private key for RSA and ECDSA. In case of a private key with passphrase an object `{ key, passphrase }` can be used (based on [crypto documentation](https://nodejs.org/api/crypto.html#crypto_sign_sign_private_key_output_format)), in this case be sure you pass the `algorithm` option.
+If `jwt.verify` is called asynchronous, `secretOrPublicKey` can be a function that should fetch the secret or public key. See below for a detailed example
 
 `options`:
 
@@ -195,6 +196,17 @@ jwt.verify(token, cert, { audience: 'urn:foo', issuer: 'urn:issuer', jwtid: 'jwt
 var cert = fs.readFileSync('public.pem'); // get public key
 jwt.verify(token, cert, { algorithms: ['RS256'] }, function (err, payload) {
   // if token alg != RS256,  err == invalid signature
+});
+
+// Verify using getKey callback
+function getKey(header, callback) {
+  // fetch secret or public key
+  const key = ...;
+  callback(null, key);
+}
+
+verify(token, getKey, options, function(err, decoded) {
+  console.log(decoded.foo) // bar
 });
 
 ```
--- a/README.md
+++ b/README.md
@@ -29,7 +29,6 @@ $ npm install jsonwebtoken
 
 `secretOrPrivateKey` is a string, buffer, or object containing either the secret for HMAC algorithms or the PEM
 encoded private key for RSA and ECDSA. In case of a private key with passphrase an object `{ key, passphrase }` can be used (based on [crypto documentation](https://nodejs.org/api/crypto.html#crypto_sign_sign_private_key_output_format)), in this case be sure you pass the `algorithm` option.
-If `jwt.verify` is called asynchronous, `secretOrPublicKey` can be a function that should fetch the secret or public key. See below for a detailed example
 
 `options`:
 
@@ -123,6 +122,7 @@ jwt.sign({
 
 `secretOrPublicKey` is a string or buffer containing either the secret for HMAC algorithms, or the PEM
 encoded public key for RSA and ECDSA.
+If `jwt.verify` is called asynchronous, `secretOrPublicKey` can be a function that should fetch the secret or public key. See below for a detailed example
 
 As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues/208#issuecomment-231861138), there are other libraries that expect base64 encoded secrets (random bytes encoded using base64), if that is your case you can pass `Buffer.from(secret, 'base64')`, by doing this the secret will be decoded using base64 and the token verification will use the original random bytes.
 
--- a/README.md
+++ b/README.md
@@ -199,13 +199,19 @@ jwt.verify(token, cert, { algorithms: ['RS256'] }, function (err, payload) {
 });
 
 // Verify using getKey callback
-function getKey(header, callback) {
-  // fetch secret or public key
-  const key = ...;
-  callback(null, key);
+// Example uses https://github.com/auth0/node-jwks-rsa as a way to fetch the keys.
+var jwksClient = require('jwks-rsa');
+var client = jwksClient({
+  jwksUri: 'https://sandrino.auth0.com/.well-known/jwks.json'
+});
+function getKey(header, callback){
+  client.getSigningKey(header.kid, function(err, key) {
+    var signingKey = key.publicKey || key.rsaPublicKey;
+    callback(null, signingKey);
+  });
 }
 
-verify(token, getKey, options, function(err, decoded) {
+jwt.verify(token, getKey, options, function(err, decoded) {
   console.log(decoded.foo) // bar
 });
 
--- a/test/verify.tests.js
+++ b/test/verify.tests.js
@@ -3,8 +3,10 @@ var jws = require('jws');
 var fs = require('fs');
 var path = require('path');
 var sinon = require('sinon');
+var JsonWebTokenError = require('../lib/JsonWebTokenError');
 
 var assert = require('chai').assert;
+var expect = require('chai').expect;
 
 describe('verify', function() {
   var pub = fs.readFileSync(path.join(__dirname, 'pub.pem'));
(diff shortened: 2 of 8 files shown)


verify.js
```diff
@@ -59,23 +59,9 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
     return done(new JsonWebTokenError('secret or public key must be provided'));
```
ziluvatar: We should move these ^ asserts inside the `getSecrect` callback too, since you could have `secretOrPublicKey` here but it could be a function that returns value / no-value later on. I mean:
```
   var hasSignature = parts[2].trim() !== '';
 	 
   if (!hasSignature && secretOrPublicKey){
   }
 	 
   if (hasSignature && !secretOrPublicKey) {
   }
```

verify.js
```diff
@@ -86,96 +72,131 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
+          return done(e);
+      }
+
+      if (!valid)
+          return done(new JsonWebTokenError('invalid signature'));
+
+      var payload;
+
+      try {
+          payload = decode(jwtString);
```
ziluvatar: 🤔 wasn't this the `decode` you wanted to clean in favor of the `decode(..., { complete: true })` call you add at the beginning of the `verify()` function?

ziluvatar: FYI Not sure why but in github if I see the diff omitting whitespaces it shows me you removed this, however, in the normal view or even if I click in View button for this file I can still see this decode call

README.md
```diff
@@ -197,6 +198,17 @@ jwt.verify(token, cert, { algorithms: ['RS256'] }, function (err, payload) {
   // if token alg != RS256,  err == invalid signature
 });
 
+// Verify using getKey callback
+function getKey(header, callback) {
```
ziluvatar: In order to use this function for JWK + openid cases we would need to add the issuer as well.

The options are:
* We add it here like `getKey(header, issuer, callback)`, and think if we want to check the issuer, if passed in options, before even calling the `getSecret` function.
* Leave it as is in this PR. Give support for that in a different one by checking the length of `getKey` function (to see if the caller expects only header and callback or header, issuer and callback)

What do you think?

README.md
```diff
@@ -29,6 +29,7 @@ $ npm install jsonwebtoken
 
 `secretOrPrivateKey` is a string, buffer, or object containing either the secret for HMAC algorithms or the PEM
 encoded private key for RSA and ECDSA. In case of a private key with passphrase an object `{ key, passphrase }` can be used (based on [crypto documentation](https://nodejs.org/api/crypto.html#crypto_sign_sign_private_key_output_format)), in this case be sure you pass the `algorithm` option.
+If `jwt.verify` is called asynchronous, `secretOrPublicKey` can be a function that should fetch the secret or public key. See below for a detailed example
```
ziluvatar: This is in wrong place, right? it should be under `jwt.verify` section, where we should indicate `secretOrPublicKey` can be a function too.

verify.js
JacoKoster: I do believe, as the other PR-owner, that the jws-package already handles these cases. But will move them anyway ;-)

verify.js
JacoKoster: You are right, should be okay now :)

README.md
JacoKoster: For the sake of progress, i would like to keep it as is. Let's make an separate issue for it and i will pick it up after this one ;-)

verify.js
```diff
@@ -43,22 +43,6 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
     return done(new JsonWebTokenError('jwt must be a string'));
   }
 
-  var parts = jwtString.split('.');
-
-  if (parts.length !== 3){
-    return done(new JsonWebTokenError('jwt malformed'));
-  }
```
ziluvatar: I think this should be kept here, it refers to the token format, not secret related, there is no need to potentially call the `getSecret` function if the jwt is a malformed

verify.js
```diff
@@ -106,28 +104,22 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
           return done(new JsonWebTokenError('invalid signature'));
 
-      var payload;
-
-      try {
-          payload = decode(jwtString);
-      } catch (err) {
-          return done(err);
-      }
+      var payload = decodedToken.payload;
```
ziluvatar: ❤️

test/verify.tests.js
```diff
@@ -67,6 +69,88 @@ describe('verify', function() {
...
+          jwt.verify(token, key, options, function (err, p) {
+              assert.isNull(err);
+              assert.deepEqual(p, payload);
+              done();
+          });
+      });
+
+      it('simple callback', function (done) {
+          var keyFunc = function(header, callback) {
+              callback(undefined, key);
```
ziluvatar: I just saw that even when you are passing the `decodedToken` the tests pass. Can you assert here you get just the content of the header from the token?

verify.js
```diff
@@ -84,98 +54,141 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
+      getSecret = secretOrPublicKey;
+  }
+  else {
+      getSecret = function(decodedToken, secretCallback) {
+          return secretCallback(null, secretOrPublicKey);
+      };
   }
 
-  return done(null, payload);
+  return getSecret(decodedToken, function(err, secretOrPublicKey) {
```
ziluvatar: 🤔 why do you pass all the decoded token now?, I think it was fine passing just the `header`, right?

JacoKoster: Because we will be using both the header and the payload of the token now

JacoKoster: Oh, wait, yes. that is wrong.

verify.js
```diff
@@ -84,98 +60,135 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
-      return done(new JsonWebTokenError('"maxAge" should be a number of seconds or string representing a timespan eg: "1d", "20h", 60'));
-    }
-    if (clockTimestamp >= maxAgeTimestamp + (options.clockTolerance || 0)) {
-      return done(new TokenExpiredError('maxAge exceeded', new Date(maxAgeTimestamp * 1000)));
-    }
-  }
-
-  return done(null, payload);
+  else {
+      getSecret = function(decodedToken, secretCallback) {
```
ziluvatar: even if it is not used, can you rename it back to `header` as well?

ziluvatar: @JacoKoster I saw your last commit after this comment, but that commit does not address this (not sure if you realized about it)

JacoKoster: Ah, yes, i forgot this one.

README.md
```diff
@@ -197,6 +198,17 @@ jwt.verify(token, cert, { algorithms: ['RS256'] }, function (err, payload) {
...
 });
 
+// Verify using getKey callback
+function getKey(header, callback) {
+  // fetch secret or public key
+  const key = ...;
+  callback(null, key);
+}
+
+verify(token, getKey, options, function(err, decoded) {
```
lbalmaceda: jwt.verify

README.md
```diff
@@ -197,6 +198,17 @@ jwt.verify(token, cert, { algorithms: ['RS256'] }, function (err, payload) {
   // if token alg != RS256,  err == invalid signature
 });
 
+// Verify using getKey callback
+function getKey(header, callback) {
+  // fetch secret or public key
```
lbalmaceda: I'd add some comment encouraging the users to cache the key with the _kid_ so they avoid fetching it multiple times.

verify.js
```diff
@@ -85,97 +62,135 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
+
+      var valid;
+
+      try {
+          valid = jws.verify(jwtString, decodedToken.header.alg, secretOrPublicKey);
+      } catch (e) {
+          return done(e);
+      }
+
+      if (!valid)
```
lbalmaceda: please be consistent on the braces usage. A few lines above this one, another one liner if clause uses them.

verify.js
```diff
@@ -85,97 +62,135 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
+          var audiences = Array.isArray(options.audience) ? options.audience : [options.audience];
+          var target = Array.isArray(payload.aud) ? payload.aud : [payload.aud];
+
+          var match = target.some(function (targetAudience) {
+              return audiences.some(function (audience) {
+                  return audience instanceof RegExp ? audience.test(targetAudience) : audience === targetAudience;
+              });
+          });
+
+          if (!match)
```
lbalmaceda: braces

verify.js
```diff
@@ -49,33 +49,10 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
-                          [ 'RS256','RS384','RS512' ] :
-                          [ 'HS256','HS384','HS512' ];
-
-  }
-
   var decodedToken;
+
   try {
-    decodedToken = jws.decode(jwtString);
+    decodedToken = decode(jwtString, { complete: true });
```
lbalmaceda: this `{complete:true}` option wasn't being passed before. How does that affect the output @ziluvatar ?

verify.js
```diff
@@ -85,97 +62,135 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
+
+      getSecret = secretOrPublicKey;
+  }
+  else {
+      getSecret = function(header, secretCallback) {
+          return secretCallback(null, secretOrPublicKey);
+      };
+  }
+
+  return getSecret(header, function(err, secretOrPublicKey) {
```
lbalmaceda: isn't enough with the `kid`? What's on the `header` that the user might want to use for key retrieval?

verify.js
JacoKoster: Noticed this as well, but because it was already on master this way. I was planning to add eslint and fixing it one go.

verify.js
ziluvatar: It's mostly the same, however, our decode implementation tries to parse always the JSON (if string), the `jws.decode` only does it when the type is `JWT`. With this and other change afterwards we reduce the amount of parsing to half.

verify.js
ziluvatar: You may want to use other headers to find the key: https://tools.ietf.org/html/rfc7515#page-10

README.md
JacoKoster: I will add https://github.com/auth0/node-jwks-rsa as an example to load the keys, that should make it more clear.

lbalmaceda:
LGTM 👨‍🎓

test/verify.tests.js
```diff
@@ -67,6 +69,90 @@ describe('verify', function() {
     });
   });
 
+  describe('secret or token as callback', function () {
```
jfromaniello: This function is not properly indented with the rest of the file and doesn't follow the newly introduced .editorconfig

JacoKoster: I'd like to make this a thing in with #486

lbalmaceda: I agree with @jfromaniello on this one. Since the bad indentation is introduced in this PR there's no need to wait for a different one to fix it.

ziluvatar:
🎉  Released on v8.3.0!! Awesome work @JacoKoster 👏

PR #491, opened 2018-06-14
Use Sinon for faking date in tests

MitMaro:
[Sinon.JS](https://github.com/sinonjs) provides handling of fake dates. This change replaces the fakeDate utility with the equivalent Sinon.JS ability.


MitMaro:
I just noticed, and not sure how I missed it the first time, that this project already has sinon as a dependency and already uses the lolex support it provides.

MitMaro:
Updated the PR to use Sinon directly.

Issue #492, opened 2018-06-14
Proposal: Test Cleanup


MitMaro:
The `test/` directory is rather difficult to follow at the moment. When adding new tests it is not always clear where the test should be added. I had some difficulties when working on #437.

I'm willing to progressively fix up the `test/` directory in a series of PRs, but since it would be a fair bit of effort, I do not want to start the process if there is not much of a chance that the PRs would be merged.

This would go nicely with the coverage report that will be added with #468.

ziluvatar:
I just merged the other PR (coverage report). Yes, indeed, the test folder is a mess, we have been added tests without thinking in real structure.

I'd like to hear your proposal before you make the effort on the individual PRs. Do you have in mind the structure you'd like to get at the end or similar thoughts?

ziluvatar:
Sorry, never mind. I didn't realize the PR you created. I'll comment there.

MitMaro:
After looking through the tests again this is what I came up with. I'm sure there are probably a few use cases that would not be captured, but they could be added later. I would add a PR for each test file created, moving any existing tests into that file as I go. 

## Claim tests

 A set of tests that are related to the claims. This would include an options for sign/verify/decode that would add a claim to the payload (expiresIn, notBefore, etc.) as well as tests that. I think these tests would use the none algorithm, and verify that the decoded token is as expected. These tests would all be independent of the algorithm used. It might be worth breaking these tests up into several files, depending on how you feel about file length. One option is to have a test file per claim, so when a new claim is added, a new file can be created. If a new claim or option that effects a claim is added, then there should only need to be tests added to this file/files.

## Algorithm tests

A set of tests that ensure that a token can be created, decoded and verified for each of the supported algorithms ('RS256', 'RS384', 'RS512', 'ES256', 'ES384', 'ES512', 'HS256', 'HS384', 'HS512', 'none'). This would include adding tests for any options that would effect the token signing. I would avoid tests that are already covered in the Claim tests, as to avoid unneeded duplication of tests. I think for simplicity, there should be four test files here, one each to handle RS*, ES*, HS* and none algorithms,

## Other tests

There are other tests related to the payload that wouldn't fit within the other test categories. I'm thinking about options like `mutatePayload` and `encoding` in particular, which effect the overall token but are independent of the claims and algorithm. In these cases a file should be added for each of these options.

## Sample file listing
```
test/claim-nbf.tests.js
    /claim-exp.tests.js
    /claim-private.tests.js
[...]

PR #493, opened 2018-06-15
Refactor sign validation tests

MitMaro:
This gives an idea of what I was thinking of for my proposal in #492. This would be the first of several PRs that I would like to create that would cleanup and complete the tests.

## Description

The validation tests for the sign method were spread across multiple files, this change brings them together into a single file. Additional tests and improvements are also made to ensure that all the validation options are fully tested.


MitMaro:
Understandable concerns, and I agree with the point. Let me try something that tackles it from the perspective of a use case. It might mean some minor overlap in the tests, but I don't think that is a bad thing in this case.

The tests in this PR should still be needed, it's just a question of the structure.

MitMaro:
I'll reject this PR, and leave a comment on the issue for this proposal.

Issue #486, opened 2018-06-08
Build pipeline


JacoKoster:
I have already 
- added a basic .editorconfig in #480 
- made a PR to upgrade to the latest Node-versions in #470 
- made a PR for adding code-coverage with Istanbul in #468 

I'd like to add two more things to make working on this repository a little bit easier
- change the jshint-config to a eslint-config
- change travis configuration to the following buildsteps (and change the package.json accordingly)
1. test
2. coverage
3. nsp
4. lint

Would this be a good idea to do?

ziluvatar:
> change the jshint-config to a eslint-config

Ok with that. Do you mean to keep both or to remove the jshint?

> change travis configuration to the following buildsteps...

Ok. Do you mean to modify the current `test` script in package.json and remove from there test + coverage + nsp + cost-of-modules and create build steps for them in travis?

About the order, probably it is better to set the lint as first step because it is faster to get a failure and avoid waiting for tests to pass to then get an error from the linter.

JacoKoster:
I fully endorse #490 as a way to solve this, i will add another PR to make sure the build breaks when doing tests.

That would change the order (indeed) to lint - test (with blocking coverage-check) - nsp

JacoKoster:
I am considering this one finished, thanks @MitMaro and @ziluvatar :-)

PR #495, opened 2018-06-19
Make code-coverage mandatory when running tests

JacoKoster:
As mentioned in #486, made the coverage-check mandatory when running test.


package.json
```diff
@@ -3,8 +3,23 @@
...
+    "statements": 95,
+    "functions": 100,
+    "branches": 95,
+    "exclude": [
+      "./test/**"
+    ],
+    "reporter": [
+      "json",
+      "lcov",
+      "text-summary"
```
ziluvatar: does this mean no HTML reporter?

JacoKoster: Preferably it should be 100% coverage, but as we are missing some tests on exception handling and @MitMaro is in the process of refactoring the tests, i was being liberate ;-)

Lcov includes a html reporter :-) 

screen shot 2018-06-25 at 20 54 35

PR #496, opened 2018-06-22
Refactor tests related to notBefore and nbf (#492)

MitMaro:
First PR for #492.

This change extracts all tests in the current files related to `notBefore` and `nbf` into a single test file. It also adds several missing related tests.

I believe I removed all references to `nbf` and `notBefore` from the current tests.

There are two TODOs in the tests, because I found a few unhandled cases that I expected would cause an error but did not. They are:

* Passing an empty string to `notBefore`. I'm not sure if this should default to `0` or should throw an error. Currently it throws a `val is not a non-empty string or a valid number. val=""` from `ms`.
* Passing `-Infinity`, `Infinity` and `NaN` to `nbf`. These all `JSON.stringify` to `null`, so I would assume should be considered invalid.

Any suggestions of missing tests are more than welcome, and I will gladly add!

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "JSON Web Token implementation (symmetric and asymmetric)",
   "main": "index.js",
   "scripts": {
-    "test": "nyc --reporter=html --reporter=text mocha && nsp check && cost-of-modules"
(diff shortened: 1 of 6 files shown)


MitMaro:
Noticed I forgot to rename the branch, I will recreate the PR.

PR #497, opened 2018-06-24
Refactor tests related to notBefore and nbf (#492)

MitMaro:
First PR for #492.

This change extracts all tests in the current files related to `notBefore` and `nbf` into a single test file. It also adds several missing related tests.

I believe I removed all references to `nbf` and `notBefore` from the current tests.

~~There are two TODOs in the tests, because I found a few unhandled cases that I expected would cause an error but did not. They are:~~

* ~~Passing an empty string to `notBefore`. I'm not sure if this should default to `0` or should throw an error. Currently it throws a `val is not a non-empty string or a valid number. val=""` from `ms`.~~
* ~~Passing `-Infinity`, `Infinity` and `NaN` to `nbf`. These all `JSON.stringify` to `null`, so I would assume should be considered invalid.~~

Any suggestions of missing tests are more than welcome, and I will gladly add!

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
   },
   "scripts": {
     "lint": "eslint .",
-    "test": "npm run lint && nyc mocha && nsp check && cost-of-modules"
+    "coverage": "nyc mocha",
+    "test": "npm run lint && npm run coverage && nsp check && cost-of-modules"
   },
   "repository": {
     "type": "git",
--- a/test/nbf.test.js
+++ b/test/nbf.test.js
@@ -1,0 +1,281 @@
+'use strict';
+
+const jwt = require('../');
+const expect = require('chai').expect;
+const sinon = require('sinon');
+const util = require('util');
+
+function base64UrlEncode(str) {
+  return Buffer.from(str).toString('base64')
+    .replace(/\=/g, "")
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_")
+  ;
+}
+
+function signWithNoBefore(payload, notBefore) {
+  const options = {algorithm: 'none'};
+  if (notBefore !== undefined) {
+    options.notBefore = notBefore;
+  }
+  return jwt.sign(payload, undefined, options);
+}
+
+const noneAlgorithmHeader = 'eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0';
+
+describe('not before', function() {
+  describe('`jwt.sign` "notBefore" option validation', function () {
+    [
+      true,
+      false,
+      null,
+      -1.1,
+      1.1,
+      -Infinity,
+      Infinity,
+      NaN,
+      ' ',
+      'invalid',
+      [],
+      ['foo'],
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -4,7 +4,8 @@
   "description": "JSON Web Token implementation (symmetric and asymmetric)",
   "main": "index.js",
   "scripts": {
-    "test": "nyc --reporter=html --reporter=text mocha && nsp check && cost-of-modules"
+    "test:unit": "nyc --reporter=html --reporter=text mocha",
```
ziluvatar: 🤔 they are not unit tests. I guess you wanted a fast way to develop the tests getting coverage feedback without the delay of nsp + cost-of-modules, right? since this is test + coverage, let's name it: `coverage` or `test:coverage` or similar.

test/nbf.test.js
```diff
@@ -0,0 +1,257 @@
...
+      true,
+      false,
+      null,
+      -1.1,
+      1.1,
+      -Infinity,
+      Infinity,
+      NaN,
+      // TODO empty string currently fails
+      // '',
```
ziluvatar: Let's create a test with the current behavior and a PR to fix this, since it should be invalid too (with same error message), and we can merge the fix on next major release.

test/nbf.test.js
```diff
@@ -0,0 +1,257 @@
...
+  describe('`jwt.sign` nbf claim validation', function () {
+    [
+      true,
+      false,
+      null,
+      undefined,
+      // TODO should these fail in validation?
+      // -Infinity,
+      // Infinity,
+      // NaN,
```
ziluvatar: As you said, yes, I think they should fail, like I said before, let's add the current behavior and a PR (same one as before?) to fix in next major.

package.json
MitMaro: Sure thing! `cost-of-modules` temporarily renames the `node_modules/` directory, and it was causing my editor to continually re-index. This was a way to run the tests without those checks.

I'm okay with a `coverage` task.

test/nbf.test.js
MitMaro: Just to make sure I understand correctly before I update:

* add a test for an empty string that checks for the `ms` error;
* add another PR that fixes the behaviour

MitMaro: Fixed in https://github.com/MitMaro/node-jsonwebtoken/commit/b037c26 before I squashed commits.

test/nbf.test.js
MitMaro: Fixed in https://github.com/MitMaro/node-jsonwebtoken/commit/b037c26 before I squashed commits.

MitMaro:
I fixed some of the formatting in https://github.com/MitMaro/node-jsonwebtoken/commit/943dab9 after fixing the issues noted above.

test/nbf.test.js
ziluvatar: Yes

PR #507, opened 2018-07-22
Refactor tests related to iat and maxAge

MitMaro:
For #492 

This change extracts all tests related to the `iat` claim and the `maxAge` verify option into two test files. Several additional tests are added that were missing from the existing tests.

Most of the references to `iat` have been removed from the existing tests. The remaining references are not directly related to the `iat` claim.

Unlike the previous test pull requests, this pull request adds some validation tests for `maxAge` to a separate file, since the option isn't directly related to a claim.

As always, suggestions or missing tests are welcome!

#### Coverage Master
```
Statements   : 97.87% ( 230/235 )
Branches     : 97.55% ( 199/204 )
Functions    : 100% ( 23/23 )
Lines        : 98.28% ( 228/232 )
```

#### Coverage Branch
```
Statements   : 97.87% ( 230/235 )
Branches     : 97.55% ( 199/204 )
Functions    : 100% ( 23/23 )
Lines        : 98.28% ( 228/232 )
```

### DIFF
--- a/test/claim-iat.test.js
+++ b/test/claim-iat.test.js
@@ -1,0 +1,273 @@
+'use strict';
+
+const jwt = require('../');
+const expect = require('chai').expect;
+const sinon = require('sinon');
+const util = require('util');
+const testUtils = require('./test-utils');
+
+const base64UrlEncode = testUtils.base64UrlEncode;
+const noneAlgorithmHeader = 'eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0';
+
+function signWithIssueAtSync(issueAt, options) {
+  const payload = {};
+  if (issueAt !== undefined) {
+    payload.iat = issueAt;
+  }
+  const opts = Object.assign({algorithm: 'none'}, options);
+  return jwt.sign(payload, undefined, opts);
+}
+
+function signWithIssueAtAsync(issueAt, options, cb) {
+  const payload = {};
+  if (issueAt !== undefined) {
+    payload.iat = issueAt;
+  }
+  const opts = Object.assign({algorithm: 'none'}, options);
+  // async calls require a truthy secret
+  // see: https://github.com/brianloveswords/node-jws/issues/62
+  return jwt.sign(payload, 'secret', opts, cb);
+}
+
+function verifyWithIssueAtSync(token, maxAge, options) {
+  const opts = Object.assign({maxAge}, options);
+  return jwt.verify(token, undefined, opts)
+}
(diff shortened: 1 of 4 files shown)


test/claim-iat.test.js
```diff
@@ -0,0 +1,174 @@
...
+const jwt = require('../');
+const expect = require('chai').expect;
+const sinon = require('sinon');
+const util = require('util');
+const testUtils = require('./test-utils');
+
+const base64UrlEncode = testUtils.base64UrlEncode;
+const noneAlgorithmHeader = 'eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0';
+
+function signWithIssueAt(issueAt, options = {}) {
```
ziluvatar: Node 4 and 5 do not support `options = {}`.

MitMaro:
I was thinking about adding async tests when I did the different algorithms. But now that you mention it, there is value in ensuring that both the async and sync calls provide similar errors.

MitMaro:
I've been on vacation this past week, and haven't had time to take a look at this PR. Now that I'm back I should get a chance.

I am going to see about adding an async test for each sync test. I will add it to this PR first, and then follow up with PRs for the other options.

MitMaro:
I've been thinking on adding async tests to the existing claim tests. I think I've come up with a solution that doesn't add a lot of complexity to each test. I will try to get this PR updated with the idea ASAP, and if it gets approved it should be easy to update the existing claim tests to follow the same format.

MitMaro:
@ziluvatar 

I've squashed https://github.com/auth0/node-jsonwebtoken/pull/507/commits/5e69a2fe997701da15ea0ebb8358869b96f717f2 into this branch.

It makes every test use the synchronous call followed by and asynchronous call, using parameterized tests to avoid duplication in the test cases.

If this pattern works for you, I will update the existing tests to follow the same pattern.

ziluvatar:
@MitMaro I merged because I liked the approach :) , however, maybe we can think in some sort of utility function that wrap every `it` with a `describe`, so if, ie, the async execution of max age fails, the reader can see right the way the problem is in the async execution and not in both which can provide a better lead for troubleshooting.

```
function syncAsyncTest(description, syncTest, asyncTest) {
  describe(description, function() {
    it('sync execution', syncTest);
    it('async execution', asyncTest);
  });
}
```
Or similar, anyway we don't need to do it now, let's see, because maybe it adds too much noise to the report without real benefit.

PR #522, opened 2018-09-11
Add complete option in jwt.verify

javespi:
Same behaviour as in `decode` method. Just return an object with the header, payload and signature when `complete: true` option is passed.

Added also a test to cover this use case.


ziluvatar:
The change is good to me, however, what is the use case for this change?
I saw the comment in the code, but if the token was already verified according to the options passed, what is the need for header + signature apart from the payload?

javespi:
Thanks @ziluvatar for the response.

My motivation to include this feature was that inside `verify` method there is already a call to `decode` passing `{ complete: true }` option:

https://github.com/auth0/node-jsonwebtoken/blob/master/verify.js#L55

My use case requires to verify the token and also returns the full JWT (header, payload, signature) of the token after verifying it. For now, I do 2 calls: one to `verify` and another one to `decode` if the verification goes OK. If we have this option in verify method I can avoid the second call to `decode` method because it's already done inside `verify`.

ziluvatar:
> My use case requires to verify the token and also returns the full JWT (header, payload, signature) after verifying it.
That's how you want to get it done, I'm asking what the use case is for the need of returning the full JWT, what do you need to do with the header or the signature? I want to understand the root reason better.

panva:
@ziluvatar there may be any number of additional properties in the JWS protected header a developer may want to access. Exposing this option would save extra processing of something which was already parsed once.

javespi:
Thanks @ziluvatar and @panva for your comments. I've already removed that comment.

test/verify.tests.js
```diff
@@ -69,6 +69,30 @@ describe('verify', function() {
     });
   });
 
+  describe('option: complete', function() {
```
ziluvatar: Sorry, I didn't realize about this before, can you follow similar approach as this option? 
https://github.com/auth0/node-jsonwebtoken/blob/master/test/option-nonce.test.js

That is: `option-complete.test.js`, using `testUtils.asyncCheck` (to report errors correctly) and  `testUtils.verifyJWTHelper` (to verify sync and async).

We are refactoring the tests moving towards that model so each option and claim has its own tests using common tooling.

README.md
```diff
@@ -138,6 +138,7 @@ As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues
 * `algorithms`: List of strings with the names of the allowed algorithms. For instance, `["HS256", "HS384"]`.
 * `audience`: if you want to check audience (`aud`), provide a value here. The audience can be checked against a string, a regular expression or a list of strings and/or regular expressions. 
   > Eg: `"urn:foo"`, `/urn:f[o]{2}/`, `[/urn:f[o]{2}/, "urn:bar"]`
+* `complete`: return an object with the decoded payload and header
```
ziluvatar:
```suggestion
* `complete`: return an object with the decoded `{ payload, header, signature }` instead of only the usual content of the payload.
```

test/verify.tests.js
javespi: Moved here https://github.com/auth0/node-jsonwebtoken/pull/522/commits/4d8f50af92b12b287f04f00e0c71dd2929ce5ba4

Apart from that, I added another unit test for `complete: false` option.

ziluvatar:
Thanks @javespi ! 🎉 Released on v8.5.0

javespi:
@ziluvatar thanks to you! Great project!

Issue #511, opened 2018-08-14
jwt.verify() not truly async.


vicary:
Despite the verify action is in fact synchronize, the callback API does not.

It is specifically this line that [released Zalgo](https://medium.com/@bluepnume/intentionally-unleashing-zalgo-with-promises-ab3f63ead2fd).

https://github.com/auth0/node-jsonwebtoken/blob/5a7fa23c0b4ac6c25304dab8767ef840b43a0eca/verify.js#L24

Which can be fixed by wrapping it as such

```js
    done = function() {
      setImmediate(callback, arguments);
    };
```

ziluvatar:
That behavior (with process.nextTick() though) was removed in this PR: https://github.com/auth0/node-jsonwebtoken/pull/302

I will leave the issue open in case the community has strong opinions about re-adding it or keep it as is. I understand the reasons, however, the pain to the event loop was already done while performing the verification synchronously, even doing that is "lying" to the consumer.

There was another discussion to remove the callback-ish style to that function, we didn't do it and now we support to get the secret/key asynchronously which gives sense to the callback now.

vicary:
Forgive my words if the opinions is too strong, but to the best of my knowledge the consensus is already there for decades.

The nature of an asynchronous API should behave like it means, implementation should be hidden behind the API even if it is a so-called lie.

The API is the contract of behavior between the library and its users, and should be followed the moment it goes public as **(Asynchronous) ...**.

IMO #302 should in fact use the **(Synchronous) ...** version without API for the desired *performance* gain, and such performance differences should be documented instead of breaking the API.

vicary:
@MitMaro thanks.

I don't have much on the callback/promise/async controversy because of the disproportionate gain of productivity, the difference in performance is tiny; I like the inline error catching in async/await though.

But the breaking of the current API should be a hotfix instead of next major, there are potential risks for people using them in production.

mcmar:
Hey, I just want to point out that if you ever re-add code to make this async (which you should do), you should also really switch from `process.nextTick` to `setImmediate`. You can read about the differences here: https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/

[...]

PR #523, opened 2018-09-15
Create and implement async/sync test helpers

MitMaro:
For #492 

In #507 there was a comment about creating a utility function to help with testing both the asynchronous and synchronous calls to `sign` and `verify`. This is the concept that I came up with based on that comment. If this is accepted I will update the existing refactored tests to use the helpers.

#### Coverage Master
```
Statements   : 97.87% ( 230/235 )
Branches     : 97.55% ( 199/204 )
Functions    : 100% ( 23/23 )
Lines        : 98.28% ( 228/232 )
```

#### Coverage Branch
```

Statements   : 97.87% ( 230/235 )
Branches     : 97.55% ( 199/204 )
Functions    : 100% ( 23/23 )
Lines        : 98.28% ( 228/232 )
```

### DIFF
--- a/test/claim-iat.test.js
+++ b/test/claim-iat.test.js
@@ -9,16 +9,7 @@ const testUtils = require('./test-utils');
 const base64UrlEncode = testUtils.base64UrlEncode;
 const noneAlgorithmHeader = 'eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0';
 
-function signWithIssueAtSync(issueAt, options) {
-  const payload = {};
-  if (issueAt !== undefined) {
-    payload.iat = issueAt;
-  }
-  const opts = Object.assign({algorithm: 'none'}, options);
-  return jwt.sign(payload, undefined, opts);
-}
-
-function signWithIssueAtAsync(issueAt, options, cb) {
+function signWithIssueAt(issueAt, options, callback) {
   const payload = {};
   if (issueAt !== undefined) {
     payload.iat = issueAt;
@@ -26,17 +17,12 @@ function signWithIssueAtAsync(issueAt, options, cb) {
   const opts = Object.assign({algorithm: 'none'}, options);
   // async calls require a truthy secret
   // see: https://github.com/brianloveswords/node-jws/issues/62
-  return jwt.sign(payload, 'secret', opts, cb);
+  testUtils.signJWTHelper(payload, 'secret', opts, callback);
 }
 
-function verifyWithIssueAtSync(token, maxAge, options) {
+function verifyWithIssueAt(token, maxAge, options, callback) {
   const opts = Object.assign({maxAge}, options);
-  return jwt.verify(token, undefined, opts)
-}
-
-function verifyWithIssueAtAsync(token, maxAge, options, cb) {
-  const opts = Object.assign({maxAge}, options);
-  return jwt.verify(token, undefined, opts, cb)
+  testUtils.verifyJWTHelper(token, undefined, opts, callback);
 }
 
 describe('issue at', function() {
@@ -53,22 +39,22 @@ describe('issue at', function() {
       {foo: 'bar'},
     ].forEach((iat) => {
       it(`should error with iat of ${util.inspect(iat)}`, function (done) {
-        expect(() => signWithIssueAtSync(iat, {})).to.throw('"iat" should be a number of seconds');
-        signWithIssueAtAsync(iat, {}, (err) => {
-          expect(err.message).to.equal('"iat" should be a number of seconds');
-          done();
+        signWithIssueAt(iat, {}, (err) => {
+          testUtils.asyncCheck(done, () => {
+            expect(err).to.be.instanceOf(Error);
+            expect(err.message).to.equal('"iat" should be a number of seconds');
+          });
         });
       });
     });
 
     // undefined needs special treatment because {} is not the same as {iat: undefined}
     it('should error with iat of undefined', function (done) {
-      expect(() => jwt.sign({iat: undefined}, undefined, {algorithm: 'none'})).to.throw(
-        '"iat" should be a number of seconds'
-      );
-      jwt.sign({iat: undefined}, undefined, {algorithm: 'none'}, (err) => {
-        expect(err.message).to.equal('"iat" should be a number of seconds');
-        done();
+      testUtils.signJWTHelper({iat: undefined}, 'secret', {algorithm: 'none'}, (err) => {
+        testUtils.asyncCheck(done, () => {
+          expect(err).to.be.instanceOf(Error);
+          expect(err.message).to.equal('"iat" should be a number of seconds');
+        });
       });
     });
   });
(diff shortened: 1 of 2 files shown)


MitMaro:
Hey @ziluvatar , just a friendly bump. I know pull requests can get lost in the weeds on GitHub.

test/claim-iat.test.js
```diff
@@ -53,22 +39,32 @@ describe('issue at', function() {
       {foo: 'bar'},
     ].forEach((iat) => {
       it(`should error with iat of ${util.inspect(iat)}`, function (done) {
-        expect(() => signWithIssueAtSync(iat, {})).to.throw('"iat" should be a number of seconds');
-        signWithIssueAtAsync(iat, {}, (err) => {
-          expect(err.message).to.equal('"iat" should be a number of seconds');
-          done();
+        signWithIssueAt(iat, {}, (err) => {
+          // the sign function swallows the error that is thrown by expect, so use done(err)
```
ziluvatar: Interesting, I didn't know about this, I guess it happens because of this: https://github.com/auth0/node-jsonwebtoken/blob/5498bdc4865ffb2ba2fd44d889fad7e83873bb33/sign.js#L176

We might tackle that at some point, for now, it's fine your handling here.

test/test-utils.js
```diff
@@ -8,6 +34,82 @@ function base64UrlEncode(str) {
...
+        callback(err);
+      }
+      else {
+        expect(syncVerified, 'Async/Sync token equality').to.deep.equal(asyncVerifiedToken);
+        callback(null, syncVerified);
+      }
+    }
+    finally {
+      if (fakeClock) {
+        fakeClock.restore();
```
ziluvatar: since the fake clock is to have the same date in both verifications, why don't we run the restore first line in this function? so that, when we call any callback the clock is the native one. Otherwise, I think we can have race conditions when the original test plays with the fake timers too 🤔

test/claim-iat.test.js
ziluvatar: Actually I just checked our current async tests, if they fail they fail with error:
`ensure "done()" is called` (because as you said sign function swallows the throw error.

test/claim-iat.test.js
```diff
@@ -53,22 +39,32 @@ describe('issue at', function() {
...
+      testUtils.signJWTHelper({iat: undefined}, 'secret', {algorithm: 'none'}, (err) => {
+        // the sign function swallows the error that is thrown by expect, so use done(err)
+        try {
+          expect(err).to.be.instanceOf(Error);
+          expect(err.message).to.equal('"iat" should be a number of seconds');
+          done();
+        }
+        catch(e) {
+          done(e);
+        }
```
ziluvatar: maybe we can move the try-catch to the helper function, otherwise all tests will need to add it.
The bad part is that if the test has a a line like the 61, it won't be clear what failed. 🤔

test/claim-iat.test.js
MitMaro: Yeah, I assumed the existing tests would have the same issue. This is definitely something I would like to fix eventually, assuming someone else doesn't get to it first.

test/test-utils.js
MitMaro: Good point on keeping the existing fake timer in the callback. Also related to the conversation, I did experiment with having multiple fake timers From what I can determine using nested calls to the Sinon fake timers is fine. The first fake timer is restored by the second fake timer's `restore` call.

I absolutely agree on having the timer restored before the callback is called.

test/claim-iat.test.js
MitMaro: I will see what I can do to break this out into a helper. Perhaps I can make it part of the `signJWTHelper` function.

test/claim-iat.test.js
MitMaro: After looking into this in further detail, this would also be a problem is #511 was every addressed. 

Long term introducing Promises would be a possible solution as it would properly handle errors thrown in an asynchronous context.

test/test-utils.js
MitMaro: Addressed in https://github.com/auth0/node-jsonwebtoken/pull/523/commits/731a76db5b78c21c73ff53a514bdb9c8081432b4 before commit was squashed.

test/claim-iat.test.js
MitMaro: Addressed in https://github.com/auth0/node-jsonwebtoken/pull/523/commits/731a76db5b78c21c73ff53a514bdb9c8081432b4 before commit was squashed.

MitMaro: I also applied the helper function to the verify tests even thought they do not use `once`. This was to keep the tests consistent and to also address this issue reappearing if the callback for verify is every made asynchronous.

test/claim-iat.test.js
```diff
@@ -53,22 +39,22 @@ describe('issue at', function() {
       {foo: 'bar'},
     ].forEach((iat) => {
       it(`should error with iat of ${util.inspect(iat)}`, function (done) {
-        expect(() => signWithIssueAtSync(iat, {})).to.throw('"iat" should be a number of seconds');
-        signWithIssueAtAsync(iat, {}, (err) => {
-          expect(err.message).to.equal('"iat" should be a number of seconds');
-          done();
+        signWithIssueAt(iat, {}, (err) => {
+          testUtils.asyncCheck(done, () => {
```
ziluvatar: It's a bit weird to use this, however, it is better than the try-catch, and more stable than simply... adding the asserts. So, good, let's keep this solution for now until we find a better one.

test/test-utils.js
```diff
@@ -1,13 +1,125 @@
...
+
+/**
+ * Correctly report errors that occur in an asynchronous callback
+ * @param {function(err): void} done The mocha callback
+ * @param {function(): void} testFunction The assertions function
+ */
+function asyncCheck(done, testFunction) {
+  try {
+    testFunction();
+    done();
```
ziluvatar: I guess we want this call (`done()`) outside of the try to avoid double called in case of error inside of it, right? (because we are passing the `done` from mocha)

ziluvatar: ^ Silly thing, I'm also fine if you leave it as is, just let me know and I'll merge.

MitMaro: Both options have different problems.

The way the code is currently means that if the `done()` call throws an error then `done(err)` will be called. This would be a double call to `done` which is generally considered a bad practice. In the case of Mocha, it will warn about the double call to `done`. Also, in this case anyways, for `done()` to fail, it would mean that Mocha is broken in some way.

By moving the `done()` call outside the `try` block then an error thrown by `done()` could be swallowed and not reported by Mocha or logged to the console.

I decided on what is there now, because I would rather Mocha tell the developer that something went wrong, then ignoring the error, or reporting that `done` was not called.

I think it's fine to keep things this way and try to fix the library later to avoid this problem altogether.

test/claim-iat.test.js
MitMaro: Long term, promises solves this problem and that appears to be direction that the JavaScript/Node.js community is moving. But that would mean converting the project to not use callbacks, which would be well outside the scope of fixing the tests.

PR #526, opened 2018-10-01
deprecate expireInMinutes and expireInSeconds - in favor of expiresIn

SansBailmare:


MitMaro:
This looks like an accidental PR or spam.

Issue #509, opened 2018-07-31
Engine spec in package.json is wrong


kastork:
The package.json says

```
  "engines": {
    "npm": ">=1.4.28",
    "node": ">=0.12"
  },
```

But the package really needs node >= 1.4 because of the use of `Object.assign` (and maybe other things).

JacoKoster:
You are right, this package currently has node 4 as the lowest version, so this should be updated.

PR #528, opened 2018-10-03
Updating Node version in Engines spec in package.json

aakash2602:
Updated Node version from 0.12 to 1.4 in package.json Engines Spec

Fixes Issue #509


ziluvatar:
I was going to check if the package is still compatible with that old version, however, I think it is better to state in the package json the versions we actually care about (we have CI set on them), so that would be `>=4` for now. Do you want to update it?

aakash2602:
where do you want me to update that in package json?

MitMaro:
I think @ziluvatar was saying to set the node version to `>=4`:

```
"node": ">=4"
```

aakash2602:
Ok will do that now

ziluvatar:
Thanks MitMaro, I was wondering what kind of release this should be, I think it will be in the next patch one, because, in the project we were really treating node 4 as the minimum node version, this is a fix just for the package.json (we could have broken compatibility with older than v4 at any moment, since it would have been fine to us).

PR #529, opened 2018-10-04
Document NotBeforeError

bploetz:
`NotBeforeError` is not documented in the list of errors that can be thrown from `verify`.


ziluvatar:
Good catch @bploetz , thanks for the PR!

PR #531, opened 2018-10-10
Fixed error message when empty string passed as expiresIn or notBefore option

andrewnester:
This PR fixes TODO items to change error message when empty string passed as `expiresIn` or `notBefore` options

### DIFF
--- a/test/claim-exp.test.js
+++ b/test/claim-exp.test.js
@@ -29,6 +29,7 @@ describe('expires', function() {
       Infinity,
       NaN,
       ' ',
+      '',
       'invalid',
       [],
       ['foo'],
@@ -42,16 +43,6 @@ describe('expires', function() {
             expect(err).to.have.property('message')
               .match(/"expiresIn" should be a number of seconds or string representing a timespan/);
           });
-        });
-      });
-    });
-
-    it(`should error with with value ''`, function (done) {
-      signWithExpiresIn('', {}, (err) => {
-        testUtils.asyncCheck(done, () => {
-          expect(err).to.be.instanceOf(Error);
-          expect(err).to.have.property('message')
-            .match(/"expiresIn" should be a number of seconds or string representing a timespan/);
         });
       });
     });
--- a/test/claim-exp.test.js
+++ b/test/claim-exp.test.js
@@ -46,12 +46,12 @@ describe('expires', function() {
       });
     });
 
-    // TODO this should throw the same error as other invalid inputs
     it(`should error with with value ''`, function (done) {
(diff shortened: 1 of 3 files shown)


andrewnester:
@MitMaro @ziluvatar may I ask you to review this change? Thanks! :)

MitMaro:
I wonder if the error message should be normalized or if this validation should be moved into the `sign_options_schema` block. But that's a decision for @ziluvatar, not I.

ziluvatar:
Good one @MitMaro , yes, probably it is better to update the schema, so if it is an string it has to have some value.
https://github.com/auth0/node-jsonwebtoken/blob/master/sign.js#L12-L13

andrewnester:
@MitMaro @ziluvatar thanks for comments! I moved validation to `sign_options_schema`

test/claim-exp.test.js
```diff
@@ -46,12 +46,12 @@ describe('expires', function() {
...
     });
 
-    // TODO this should throw the same error as other invalid inputs
     it(`should error with with value ''`, function (done) {
       signWithExpiresIn('', {}, (err) => {
         testUtils.asyncCheck(done, () => {
           expect(err).to.be.instanceOf(Error);
-          expect(err).to.have.property('message', 'val is not a non-empty string or a valid number. val=""');
+          expect(err).to.have.property('message')
+            .match(/"expiresIn" should be a number of seconds or string representing a timespan/);
```
MitMaro: You should be able to move this test above into the array in `jwt.sign "expiresIn" option validation` block.

test/claim-nbf.test.js
```diff
@@ -51,7 +51,8 @@ describe('not before', function() {
       signWithNotBefore('', {}, (err) => {
         testUtils.asyncCheck(done, () => {
           expect(err).to.be.instanceOf(Error);
-          expect(err).to.have.property('message', 'val is not a non-empty string or a valid number. val=""');
+          expect(err).to.have.property('message')
+            .match(/"notBefore" should be a number of seconds or string representing a timespan/);
```
MitMaro: Ditto to above, you should be able to move this test above into the array in `jwt.sign "expiresIn" option validation` block.

andrewnester:
@MitMaro indeed, good point, thanks for the comments! Updated in latest commit.

andrewnester:
@ziluvatar sorry for bothering again, any other comments on this PR? :)

ziluvatar:
@andrewnester nothing to sorry about, thanks for the interest! I will merge this one since we were returning an error anyway, however, in this case it will return a better known descriptive error.

PR #540, opened 2018-10-20
Add verify option for nonce validation

kazuki229:
I want to verify nonce value in jwt.
Then I found the following pull request.
https://github.com/auth0/node-jsonwebtoken/pull/344

But because tests and README have not been added, it has not yet been merged.
So I did add tests and README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues
 * `clockTolerance`: number of seconds to tolerate when checking the `nbf` and `exp` claims, to deal with small clock differences among different servers
 * `maxAge`: the maximum allowed age for tokens to still be valid. It is expressed in seconds or a string describing a time span [zeit/ms](https://github.com/zeit/ms). Eg: `1000`, `"2 days"`, `"10h"`, `"7d"`. A numeric value is interpreted as a seconds count. If you use a string be sure you provide the time units (days, hours, etc), otherwise milliseconds unit is used by default (`"120"` is equal to `"120ms"`).
 * `clockTimestamp`: the time in seconds that should be used as the current time for all necessary comparisons.
-* `nonce`: if you want to check nonce, provide a value here.
+* `nonce`: if you want to check `nonce` claim, provide a string value here. It is used on Open ID for the ID Tokens. ([Open ID implementation notes](https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes))
 
 
 ```js
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues
 * `clockTolerance`: number of seconds to tolerate when checking the `nbf` and `exp` claims, to deal with small clock differences among different servers
 * `maxAge`: the maximum allowed age for tokens to still be valid. It is expressed in seconds or a string describing a time span [zeit/ms](https://github.com/zeit/ms). Eg: `1000`, `"2 days"`, `"10h"`, `"7d"`. A numeric value is interpreted as a seconds count. If you use a string be sure you provide the time units (days, hours, etc), otherwise milliseconds unit is used by default (`"120"` is equal to `"120ms"`).
 * `clockTimestamp`: the time in seconds that should be used as the current time for all necessary comparisons.
+* `nonce`: if you want to check nonce, provide a value here.
 
 
 ```js
--- a/test/option-nonce.test.js
+++ b/test/option-nonce.test.js
@@ -3,6 +3,7 @@
 const jwt = require('../');
 const expect = require('chai').expect;
 const util = require('util');
+const testUtils = require('./test-utils')
 
 describe('nonce option', function () {
   let token;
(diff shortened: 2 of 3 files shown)


test/option-nonce.test.js
```diff
@@ -0,0 +1,50 @@
...
+      nonce: 'abcde',
+    },
+  ].forEach((testCase) => {
+    it(testCase.description, function (done) {
+      expect(jwt.verify(token, undefined, { nonce: testCase.nonce })).to.not.throw;
+      jwt.verify(token, undefined, { nonce: testCase.nonce }, (err) => {
+        expect(err).to.be.null;
+        done();
+      })
+    });
```
ziluvatar: Can you do the tests using the helpers that @MitMaro prepared?
It will internally assert sync and async behavior (like you did well here, just for consistency across tests)

verify.js
```diff
@@ -179,6 +179,12 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
       }
     }
 
+    if (options.nonce) {
```
ziluvatar: Could you return error if `options.nonce` is not a string?

README.md
```diff
@@ -139,6 +139,7 @@ As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues
 * `clockTolerance`: number of seconds to tolerate when checking the `nbf` and `exp` claims, to deal with small clock differences among different servers
 * `maxAge`: the maximum allowed age for tokens to still be valid. It is expressed in seconds or a string describing a time span [zeit/ms](https://github.com/zeit/ms). Eg: `1000`, `"2 days"`, `"10h"`, `"7d"`. A numeric value is interpreted as a seconds count. If you use a string be sure you provide the time units (days, hours, etc), otherwise milliseconds unit is used by default (`"120"` is equal to `"120ms"`).
 * `clockTimestamp`: the time in seconds that should be used as the current time for all necessary comparisons.
+* `nonce`: if you want to check nonce, provide a value here.
```
ziluvatar:
```suggestion
* `nonce`: if you want to check `nonce` claim, provide a string value here. It is used on Open ID for the ID Tokens. ([Open ID implementation notes](https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes))
```

MitMaro:
Kudos to @kazuki229 for adding tests!

I think there are a few additional tests that could be easily added. I will take a look this weekend and provide some direction to @kazuki229 .

verify.js
MitMaro: With the above validation is added, then we should test that the following values result in the error being returned:

```
[
  true,
  false,
  null,
  -1,
  0,
  1,
  -1.1,
  1.1,
  -Infinity,
  Infinity,
  NaN,
  '',
  ' ',
  [],
  ['foo'],
  {},
  {foo: 'bar'},
]
```

test/option-nonce.test.js
```diff
@@ -0,0 +1,50 @@
...
+
+describe('nonce option', function () {
+  let token;
+
+  beforeEach(function () {
+    token = jwt.sign({ nonce: 'abcde' }, undefined, { algorithm: 'none' });
+  });
+  [
+    {
+      description: 'should work with string',
```
MitMaro: Minor item: `'should work with a string'` instead?

test/option-nonce.test.js
```diff
@@ -0,0 +1,50 @@
...
+  });
+  [
+    true,
+    'invalid',
+    [],
+    ['foo'],
+    {},
+    { foo: 'bar' },
+  ].forEach((nonce) => {
+    let tokenhoge = jwt.sign({ foo: 'bar' }, undefined, { algorithm: 'none' });
```
MitMaro: You should be able to remove this line and reuse the `token` generated in the `beforeEach` above.

test/option-nonce.test.js
MitMaro: @kazuki229 Some direction that hopefully will make things a little easier.

Add `const testUtils = require('./test-utils');` to get the test utilities.

To use the test helpers for `verify`:

```
testUtils.verifyJWTHelper(token, undefined, {nonce: testCase.nonce}, (err, decoded) => {
  testUtils.asyncCheck(done, () => {
    expect(err).to.be.null;
    expect(decoded).to.have.property('nonce', 'abcde');
  });
});
```

To check for an error you would do something along the lines of:

```
testUtils.verifyJWTHelper(token, undefined, {nonce}, (err) => {
  testUtils.asyncCheck(done, () => {
    expect(err).to.be.instanceOf(jwt.JsonWebTokenError);
    expect(err).to.have.property('message', 'nonce must be a non-empty string')
  });
});
```

Ping me if you need any other assistance and I will gladly help!

kazuki229:
@ziluvatar @MitMaro 
I'm sorry too late.
I fixed the points pointed out, So please review again.:bow:

ziluvatar:
Thanks @kazuki229 !! Good job 👏 I'll create a minor version with this and some other patch changes that are on master.

ziluvatar:
Released on v8.4.0 🎉

kazuki229:
@ziluvatar 
Thank you for your quick response! :tada::tada:

PR #547, opened 2018-10-26
Map different issuers to specific keys in `verify`

### LINKED ISSUE #238: Support validating with different public keys based on issuer name
It would be useful to be able to give each issuer it's own public and private key pair, and when validating, use the issuer field to determine which key to attempt validation with.

For example, if I have two issuers, `foo` and `bar`, an additional config option like this:

```
jwt.verify(token, {
  foo: fs.readFileSync('foo.public.key'),
  bar: fs.readFileSync('bar.public.key'),
});
```

Where verify validated against the public key for the issuer specified in `token`.

jgierer12:
Closes #238

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ jwt.sign({
 `secretOrPublicKey` is a string or buffer containing either the secret for HMAC algorithms, or the PEM
 encoded public key for RSA and ECDSA.
 If `jwt.verify` is called asynchronous, `secretOrPublicKey` can be a function that should fetch the secret or public key. See below for a detailed example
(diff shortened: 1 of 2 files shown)


ziluvatar:
I see the use for this, however, if you need this it is better to extend our current implementation to select the key, currently `getSecret()` only receives the `header`, we could modify it so in case the `secretOrPublicKey` is a function with 3 arguments we also provide the payload.

https://github.com/auth0/node-jsonwebtoken/blob/master/verify.js#L80

That way in your function you select the right key based on issuer (or issuer + kid).

PR #549, opened 2018-11-01
Removing someone else's .editorconfig and ignoring it.

dabockster:
Someone else included their editor config file by accident. Removing it and ignoring it.

Let's also try not to get into the spaces vs tabs argument again. I respect both options. 👍

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,0 @@
-[*]
-indent_style = space
-indent_size = 2
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 .DS_Store
 .nyc_output
 coverage
+.editorconfig


Macil:
Doesn't this project use 2-space indents in its source files? I don't mind whether the project uses spaces or tabs, but ideally it only uses one or the other, and it makes sense for it to have an .editorconfig file present so people's editors that support it can auto-configure to match whatever style the project's maintainers want. (.editorconfig files don't carry user-specific data in them like many IDE project files do; they're built to be committed.)

MitMaro:
I don't think this file should be removed as my editor uses it to ensure consistency in this project. If anything I think it should be expanded to ensure a final newline and to enforce proper newlines. `root` should also be added to ensure that some other user level configuration doesn't influence the configuration.

```
root = true

[*]
indent_style = space
indent_size = 2
end_of_line = lf
insert_final_newline = true
```

ziluvatar:
I agree with last two comments, that file helps to develop this library with consistent format.
I think we can close the PR.
Thanks anyway @dabockster for thinking about improving the repo maintenance :)

PR #555, opened 2018-12-02
Add tests for private claims in the payload

MitMaro:
Another PR for #492

This change adds tests for private claims added to the payload during sign and ensures that after verifying the payload contains the expected claim. There are not currently any dedicated tests to private claims, so these test cases are new.

#### Coverage Master
```
Statements   : 96.72% ( 236/244 )
Branches     : 97.21% ( 209/215 )
Functions    : 100% ( 23/23 )
Lines        : 97.1% ( 234/241 )
```

#### Coverage Branch

```
Statements   : 96.72% ( 236/244 )
Branches     : 97.21% ( 209/215 )
Functions    : 100% ( 23/23 )
Lines        : 97.1% ( 234/241 )
```

### DIFF
--- a/test/claim-private.tests.js
+++ b/test/claim-private.tests.js
@@ -1,0 +1,73 @@
+'use strict';
+
+const expect = require('chai').expect;
+const util = require('util');
+const testUtils = require('./test-utils');
+
+function signWithPayload(payload, callback) {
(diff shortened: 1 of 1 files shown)


MitMaro:
This is in part just testing that `JSON.stringify` works as expected, so I don't know how much value that these tests add. The only thing I can think of is that It does ensure that these inputs remain stable in new releases.

PR #563, opened 2018-12-30
Adding scope in options

JiteshShivnani:
scope is added in option which can be validated against token payload. This is a non mandatory parameter. Supporting test cases are also added.

### DIFF
--- a/sign.js
+++ b/sign.js
@@ -20,7 +20,8 @@ var sign_options_schema = {
   jwtid: { isValid: isString, message: '"jwtid" must be a string' },
   noTimestamp: { isValid: isBoolean, message: '"noTimestamp" must be a boolean' },
   keyid: { isValid: isString, message: '"keyid" must be a string' },
-  mutatePayload: { isValid: isBoolean, message: '"mutatePayload" must be a boolean' }
+  mutatePayload: { isValid: isBoolean, message: '"mutatePayload" must be a boolean' },
+  scope: { isValid: isString, message: '"scope" must be a string' }
 };
 
 var registered_claims_schema = {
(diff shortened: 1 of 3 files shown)


JiteshShivnani:
@ziluvatar can you please review this? Also there is a security vulnerability in master branch.

ziluvatar:
Few things:
* If we add that validation on signing that is a breaking change, because we could have consumers currently creating JWTs with other formats in `scope` claim and updating this change would break their integration.
* I'm not sure if we should open the box of "let's put validations of arbitrary fields on the library", this library provides validations over the claims on the JWT spec + nonce (that I'm now thinking we shouldn't have added it without further consideration). This library should be the base where other libraries for specific JWT usages could add their specific logic on top (OpenID, authorization, etc)

panva:
IANA does not have `scope` registered from any specification. Adding validations for such a thing would therefore be proprietary.

To @ziluvatar's point, `nonce` definitely has a place, its type is defined and it is registered.

If I may suggest, instead of adding each specific value as verifiable like this PR suggests, why not either

- do this after getting the verified payload
- allow for literally any claim to be custom-validated with something like 


```js
jsonwebtoken.verify(jwt, key, {
  custom(header, payload) {
    // custom code verifying here.
  }
});
```

ziluvatar:
Agree with providing some kind of custom validations, we could talk about that in a different PR.
About the IANA registration, even if registered, I think it is out of scope of this library. We could mix the "custom validation" interface with some new lib-friend where we can add different validations based on the different jwt profiles.

@JiteshShivnani closing the PR, if you want to dig into the custom validation interface be free to open a different PR.

PR #577, opened 2019-02-14
Force use_strict during testing

lionello:
Our project was using JS strict mode and we got errors like
```
TypeError: Cannot create property 'iat' on string
```

Enabling `--use_strict` during the mocha test revealed the culprit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "lint": "eslint .",
-    "coverage": "nyc mocha",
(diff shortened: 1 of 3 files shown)


ziluvatar:
Could you add some tests on `claim-iat.test.js` file to assert that behavior is not changing for string payloads (proper stringified json and just a plain string like in you issue)

lionello:
@ziluvatar Done. Please check if it's what you had in mind.

ziluvatar:
Thanks @lionello 👏

PR #597, opened 2019-05-04
Dist files for web

gustavosatheler:


panva:
@gustavojucoskifernandes Although you might be able to use `jsonwebtoken` in a browser through transpiled code using e.g. webpack this library's focus is the node.js platform and we lack the test suite to confirm it all works as intended, we won't be focusing on the browser platform and would recommend use of JWT modules that utilize the browser built-in [`Web Cryptography API`](https://www.w3.org/TR/WebCryptoAPI/).

PR #598, opened 2019-05-13
Added option for maxExpiration in verify

Sherex:
Adds the option to specify maxExpiration in the verify function.
It will require the token to include exp and iat in the payload, and exp has to be less than what is specified in maxExpiration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues
 * `clockTolerance`: number of seconds to tolerate when checking the `nbf` and `exp` claims, to deal with small clock differences among different servers
 * `maxAge`: the maximum allowed age for tokens to still be valid. It is expressed in seconds or a string describing a time span [zeit/ms](https://github.com/zeit/ms). 
   > Eg: `1000`, `"2 days"`, `"10h"`, `"7d"`. A numeric value is interpreted as a seconds count. If you use a string be sure you provide the time units (days, hours, etc), otherwise milliseconds unit is used by default (`"120"` is equal to `"120ms"`).
+* `maxExpiration`: the maximum time tokens are allowed to expire in. It is expressed in seconds or a string describing a time span [zeit/ms](https://github.com/zeit/ms).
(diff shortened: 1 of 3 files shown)


panva:
@Sherex thank you for taking the time to open a good looking PR. I'm wondering tho - isn't this new option to the same end as the `maxAge` option? The only difference is when one of these throws.

Yours looks more like content validation to make sure it follows policy, where `maxAge` just throws when the policy is over reached regardless of the actual `exp`?

Sherex:
@panva Yeah that was the point, I felt it fitted more for an API solution, because the user has to know about the limit (or get an error message). That may avoid an unexpected time-out.

But it may also create more problems than it's worth. For example, if the server sets `maxExpiration` down 1 minute, then it may break other clients using that API...

And the user can't use a token which lasts forever, I feel like it's a tad more secure(?), but it might just be overcomplicating things.

I also didn't want to use the `maxAge`, because it was marked as deprecated in the .ts file.
Do you know if that's a bug?

[DefinitelyTyped/jsonwebtoken](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/efa6b9e11eb0167547d6611490be48b94bc8d7ea/types/jsonwebtoken/index.d.ts#L75)

Npm module:
[@types/jsonwebtoken](https://www.npmjs.com/package/@types/jsonwebtoken)

panva:
> And the user can't use a token which lasts forever

`maxAge` is perfectly suited for that as well

Since the only thing that changes is WHEN an error is thrown i don't think this PR is needed, it would introduce just another way of doing what's already possible. Unlike the other options `exp` is a sliding target.

> I also didn't want to use the `maxAge`, because it was marked as deprecated in the .ts file.
> Do you know if that's a bug?

I don't know why the types say this is deprecated. We're not the maintainers of those types and it does look like a bug with the types, nothing in the code seems to suggest the option is deprecated.

Sherex:
Yups, I agree, probably best to leave it as is. :)

Thanks for the review!

panva:
thanks @Sherex

PR #639, opened 2019-10-11
Some fixes to token decoding

jsumners:
### Description

This PR tweaks a couple of items in the decoding of web tokens.

### Checklist

### DIFF
--- a/decode.js
+++ b/decode.js
@@ -1,4 +1,5 @@
 var jws = require('jws');
+var parse = require('secure-json-parse').parse;
 
 module.exports = function (jwt, options) {
   options = options || {};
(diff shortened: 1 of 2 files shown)


panva:
@jsumners the change is lacking test coverage and also, can you provide more background the change in general?

jsumners:
I suppose a direct test for checking if `isObject` picks up `[]` or `42` could exercise this PR. But it'd ultimately be checking a well known JavaScript pattern so I don't see much utility in it.  Otherwise, I would hope the existing test suite already covers this.

More background would require using a webform and process I don't want to deal with.

PR #645, opened 2019-11-24
fixes numeric expiresIn sign option by normalizing handling in timest…

jduarter:
Documentation says a number of seconds before expiration can be specified in the expiresIn option of the sign() function. Nevertheless, function was handling numeric values as milliseconds. 

This commit multiplies input by 1000 to fix this issue and make the library act as documented.

### Checklist

- [NA] I have added documentation for new/changed functionality in this PR or in auth0.com/docs


jduarter:
I will be analyzing this more extensively as a lot of tests seems to be failing after my change. I will open a new PR after I finish with it.

Issue #458, opened 2018-04-07
Key rotation implementation


Extarys:
I looked but it seems that this feature is missing or I didn't find it in any documentation.

I was starting to use HS512 but then realized it would be better for a multi-server setup to use RSA. My issue is that if I want to change the key every x months, all old JWT will not be valid anymore. Is there a way to put them in an array or something and JWT would always use the last one to encrypt and sign, and if the validation failed it will try the last key and re-sign with the new one while marking the old one as revoked?

thanks!

PR #648, opened 2019-12-03
Support key rotation: let verify secretOrPublicKey be an array

quintinm-dev:
### Description
@hmcfletch and I would find it useful to have `jwt.verify` also accept an array of keys, such that we can rotate our signing key & pass both keys to verify during a grace period. It'd be great to hear your feedback if it's something you're interested in.

We wrapped the existing `secretOrPublicKey` argument into an array rather than adding another `options.algorithms` field, as that seemed a bit more graceful.

It will:
- error if unsigned token given with secretOrPublicKey array containing truthy values
- error if signed token given with secretOrPublicKey array not containing truthy values
- error if token algorithm is unsupported by default algorithms of every key (if options.algorithms is unset)
- then successfully decode if any key verifies the token

### References
[Key rotation previously discussed here](https://github.com/auth0/node-jsonwebtoken/issues/458) (JWK seemed like overkill for our use case)

### Testing

We put the basic functionality tests at the bottom of `verify.tests.js`. We put more extensive tests where we mixed in different algorithms in a separate file.

### Checklist

Docs can be added if there's interest in merging.


Ry7n:
This is code I've written half a dozen times now, and it would be nice if I didn't have to again, although I worry that accepting `string | string[]` will [incur unwanted cost](https://alligator.io/js/v8-engine/#fix-function-argument-types) in what is a performance-sensitive part of the code (since many services will call this on every request).

QMuncaster:
Good point. I was writing with readability in mind rather than performance, when maybe it should've been the opposite. For example, I duplicate work to get the default supported algorithms (once to determine if any keys support the token algorithm, and then again to check if a specific key supports the algorithm before verifying), and I loop over the key array 3 separate times. This could all be done in a single loop, maybe at the cost of readability.

I'm not sure if the argument polymorphism would be a huge concern. We use `Array.isArray` on the `secretOrPublicKey` argument - does that qualify as a [dynamically bound operation](https://mrale.ph/blog/2015/01/11/whats-up-with-monomorphism.html)? I could probably look into how many inline cache entries are created when a user calls verify with `secretOrPublicKey` as a function, a string, a buffer, or various arrays.

quintinm-dev:
Closing due to inactivity

PR #651, opened 2019-12-17
chore: update jws

panva:
### Description

`jsonwebtoken` does its own JWS algorithm check which didn't suffer the flaw of the `jws`/`jwa` - it would still be nice to be on the most recent semver major line.

### References

https://github.com/brianloveswords/node-jws/compare/v3.2.2...v4.0.0
https://github.com/brianloveswords/node-jwa/compare/v1.4.1...v2.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "url": "https://github.com/auth0/node-jsonwebtoken/issues"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


ziluvatar:
It is still possible to pass a list of algorithms (even if we don't formally accept them in the Readme).
I'll mark this for next-major.

PR #660, opened 2020-01-15
feat: expose timespan as convenience

jiverson:
### Description

Expose `timespan` as an utility function so as to be able to expose or return the `expiresIn` from a signed token. Just a simple helper function to not have to decode the token to determine when the token expires. As well as having access to `ms`.

For example, many web based applications only do a silent refresh of the token when receiving a 401. A more better experience is the client to know when the token will expire and proactively refresh. While it is true you can decode the token on the client but that requires another unneeded lib or another step in the flow. Ideally just pass back the expiration date with the token.

I agree there are other ways of doing this but it just seems like a very simple and valid way of ensuring the code cohesiveness.

### References

```javascript
const { sign, timespan } = require("jsonwebtoken")

exports.createAccessToken = ({ id: userId }) => {
    const exp = timespan("15m")

    return {
        token: sign({ userId, exp }, process.env.ACCESS_TOKEN_SECRET),
        expiresIn: exp
    }
}
```

### Testing

None that I am aware of.

### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@ module.exports = {
   decode: require('./decode'),
   verify: require('./verify'),
   sign: require('./sign'),
-  timespan: require('./timespan'),
+  timespan: require('./lib/timespan'),
   JsonWebTokenError: require('./lib/JsonWebTokenError'),
   NotBeforeError: require('./lib/NotBeforeError'),
   TokenExpiredError: require('./lib/TokenExpiredError'),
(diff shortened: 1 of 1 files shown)


ziluvatar:
Technically there is no issue with this, however, I don't see it related to JWT business and will increase the public api surface to maintain.

Also `exp` value is not a duration value, it is a moment during time defined in seconds.

If you want to have there the `exp` value without decoding you can use the option `mutatePayload`, like:
https://runkit.com/embed/4nxcdxz2k789

```
const jwt = require("jsonwebtoken")
const ms = require("ms")

const payload = { data: true };
const token = jwt.sign(payload, 'some-secret', { mutatePayload: true, expiresIn: '15m' });

console.log(payload.exp)
```

PR #664, opened 2020-01-17
Allow passing json option to verify

LinusU:
### Description

The `decode` function already accepts a `json` option, which when set to true will always try and parse the resulting payload as JSON, and throw an error if that fails.

This patch adds the same option to the `verify` function, by passing the `json` option thru to the `decode` function.

### Testing

This can be tested by passing `{ json: true }` to `verify`, with a token that has a string payload. That should throw a JSON parse error.

### Checklist

### DIFF
--- a/verify.js
+++ b/verify.js
@@ -66,7 +66,7 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
   var decodedToken;
 
   try {
-    decodedToken = decode(jwtString, { complete: true });
+    decodedToken = decode(jwtString, { complete: true, json: options.json });
   } catch(err) {
     return done(err);
   }


ziluvatar:
The parsing [is always done](https://github.com/auth0/node-jsonwebtoken/blob/master/decode.js#L10) if the content was a string. There is no `json` option on `decode`.

Maybe you were looking for `complete: true` option to get all the parsed parts of a token?

LinusU:
@ziluvatar actually, the code in the library will try and parse it and fallback to giving the string back if the parsing fails. Passing `json: true` in will guarantee that the value is parsed as JSON, and `throw` an error if the parsing fails.

The `json: true` option is sent to the `jws` library here: https://github.com/auth0/node-jsonwebtoken/blob/a6235fa561b5c30884c97ea0b30c3db3b546ae2c/decode.js#L5

and the `jws` library than looks for that option here: https://github.com/brianloveswords/node-jws/blob/1389f6d3d018b2da4af0167f48e44f90d811ae87/lib/verify-stream.js#L70-L71

I would appreciate if we could reopen this while we are discussing this so that it doesn't get lost ☺️ thanks!

ziluvatar:
Ok, I see what you mean now. Yeah, I think it makes sense. If the token header contains `type: 'JWT'` the current code would fail already, however, by spec the presence of that header is optional.

We could have a `jsonPayload` option added to `verify()` and `decode()` functions. If enabled the library would return an error in similar way it does with other verifications.

Thoughts? Do you want to implement its support together with tests and readme update?

crrobinson14:
This PR is ancient - is there a reason it was never merged? Right now the return type of `verify()` is `string | Jwt` because there's no way to enforce JSON payloads. This means you always need a guard like this:

``` 
if (typeof decodedToken === 'string') {
    warn('Session: Unable to validate token: invalid payload', decodedToken);
    throw new AccessDeniedError(AUTH_ERROR);
  }
```

BUT this is really hard to test in most applications because it's hard to even generate sample tokens to test this kind of code in most JWT debuggers, so this code almost always gets an `istanbul-ignore-next` flag, which I hate to see in such a critical code block. Since string payloads are basically unheard-of / useless in major JWT-based auth products and flows, having a more strictly-typed verification method that only tolerate JSON payloads (I mean... it's a JSON Web Token...) would be nice to have.

PR #717, opened 2020-04-28
Add mutateHeader option

### LINKED ISSUE #714: Header option should be used without any modification
### Describe the problem you'd like to have solved

I'm trying sign a JWT token with the header:
```javascript
const token = jwt.sign(payload, secret, {
  header: {
    typ: "JWT",
    alg: "HS256",
  }
});
```

But inside the `jsonwebtoken`, the header option is assigned with the default header values:
```javascript
var header = Object.assign({
  alg: options.algorithm || 'HS256',
  typ: isObjectPayload ? 'JWT' : undefined,
  kid: options.keyid
}, options.header);
```

This behavior invert the typ and alg position and add a kid property to header:
```javascript
{ alg: "HS256", typ: "JWT", kid: undefined }
```

### Describe the ideal solution
I think when `options.header` is provided, it should be used without any modifications

weslenng:
### Description

The mutateHeader option allows you to control the current assignment which occurs between `options.headers` and the default header values (alg, typ and kid). Default behaviour is unchanged since mutatePayload defaults to true.

Closes #714

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ encoded private key for RSA and ECDSA. In case of a private key with passphrase 
 * `header`
 * `keyid`
 * `mutatePayload`: if true, the sign function will modify the payload object directly. This is useful if you need a raw reference to the payload after claims have been applied to it but before it has been encoded into a token.
-
+* `mutateHeader`: if false, the sign function will not modify the `options.header` by assigning with the default headers values.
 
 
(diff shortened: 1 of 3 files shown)


jhiner:
As https://github.com/auth0/node-jsonwebtoken/issues/714, closing this as well.

PR #722, opened 2020-05-17
Use native Array.includes() instead of lodash.includes

dandv:
### Description

This PR replaces the one occurrence of `lodash.includes` with the native [Array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes), which has been supported in Node since v6, now End Of Lifed.

### References

For a security library,
* It's valuable to reduce the number of dependencies. This is a step towards modernizing the codebase.
* The target audience keeps their Node version current. Node v6 was released in 2016 and [EOLed more than a year ago](https://web.archive.org/web/20190317014458/https://nodejs.org/en/about/releases/). Thus the use of native Array.includes is supported by all active Node releases.

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,16 +17,6 @@ commands:
           command: npm test
 
 jobs:
-  node-v4:
-    docker:
-      - image: node:4
-    steps:
-      - test-nodejs
-  node-v5:
-    docker:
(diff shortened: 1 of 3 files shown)


eknowles:
looks reasonable 👍

PR #738, opened 2020-08-12
Rename "invalid token" error to "jwt malformed"

jmacmahon:
# Description

This brings the message for a token with (for example) a trailing space into line with the listed potential error messages in the README.

"invalid token" is confusing and not a listed possible error message -- is it invalid because it was corrupt, unauthorized, expired, etc.?

This also adds a test for the existing case where "jwt malformed" is returned, which was otherwise untested.

### Checklist

### DIFF
--- a/test/jwt.hs.tests.js
+++ b/test/jwt.hs.tests.js
@@ -96,16 +96,26 @@ describe('HS256', function() {
     });
   });
 
-  describe('should fail verification gracefully with trailing space in the jwt', function() {
+  describe('should fail verification gracefully on malformed token', function() {
     var secret = 'shhhhhh';
     var token  = jwt.sign({ foo: 'bar' }, secret, { algorithm: 'HS256' });
 
-    it('should return the "invalid token" error', function(done) {
+    it('should return the "jwt malformed" error with a trailing space', function(done) {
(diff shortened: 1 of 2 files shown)


jmacmahon:
Any thoughts on this?

verify.js
```diff
@@ -72,7 +72,7 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
   }
```
xayrulloergashev, line 72: asasasASASA

test/jwt.hs.tests.js
```diff
@@ -96,16 +96,26 @@ describe('HS256', function() {
     });
   });
 
-  describe('should fail verification gracefully with trailing space in the jwt', function() {
+  describe('should fail verification gracefully on malformed token', function() {
     var secret = 'shhhhhh';
     var token  = jwt.sign({ foo: 'bar' }, secret, { algorithm: 'HS256' });
 
-    it('should return the "invalid token" error', function(done) {
+    it('should return the "jwt malformed" error with a trailing space', function(done) {
```
DevLereko, line 103: What might have happened to have the malformed JWT, it's because on my app, it has been working fine, and then lately I get `jwt malformed` error message, if I refresh the page, I will no longer see that error again, can you please give me details on this.

PR #759, opened 2021-01-12
Civ 2498 use crypto manager for all ec

jpsantosbh:
We want to use the cripto manager in more than just ES256K support.
This is required for our Cripto Utils lib. Since now for all key types the private key is managed, and never is exported.


panva:
I've closed this PR because it was most likely opened either by accident or against the wrong origin.

PR #793, opened 2021-10-04
Don't error out on undefined options

### LINKED ISSUE #703: Cannot pass undefined for optional options
### Description

When signing/verifying with options, a present attribute is not allowed to be undefined, which results in some unexpected an obnoxious code gymnastics when passing optional configured values through to the library.

There is a very strong precedent in JavaScript libraries that an explicitly-provided `undefined` value is acceptable for an optional value, but this library does not accept that.

### Reproduction

```js
const jwt = require('jsonwebtoken');

// Assume the config is obtained externally and not inline.
const config = {
    secret: 'foo',
    audience: 'urn:example'
};

console.log(jwt.sign(
    {},
    config.secret,
    { issuer: config.issuer, audience: config.audience }
));
```

Expected behavior: a valid token is logged.

Actual behavior: an error is thrown (`Error: "issuer" must be a string`).

The simplest workaround is to mutate the object by removing keys that exist but are undefined:

```js
const jwt = require('jsonwebtoken');

function removeUndefined(obj) {
    Object.keys(obj).forEach(key => {
        if (obj[key] === undefined) {
            delete obj[key];
        }
    });

    return obj;
}

// Assume the config is obtained externally and not inline.
const config = {
    secret: 'foo'
};

console.log(jwt.sign(
    {},
    config.secret,
    removeUndefined({ issuer: config.issuer, audience: config.audience })
));
```

However, this should not be required in my opinion. Validation should skip optional keys whose value is `undefined`.

### Environment

- **Version of this library used:** 8.5.1
- **Version of the platform or framework used, if applicable:** N/A
- **Other relevant versions (language, server software, OS, browser):** N/A (bug is not platform-specific)
- **Other modules/plugins/libraries that might be involved:** None

KuSh:
### Description

This allow calling sign with undefined options without erroring out

Fixes #703 

### Checklist


iamart (approved):

iamart:
Wondering why those tests were even created.

PR #801, opened 2021-12-28
Remove node version checks

RDIL:
### Description

This is a breaking change - it removes the support for node 4-7, since those versions are very old and not widely used anymore.

Pros:

- Removes the dependency on semver
- Removes the splice calls
- Slight reduction of code size

Cons:

- For the few people out there still using these versions, they will be forced to either update node or stay on an old library version.

If the maintainers aren't interested in dropping support for these old versions, that's fine, I just think that this change is (while breaking), harmless enough, because I really don't know of anybody who is still using node 4, 5, 6, or 7.

### References

None really.

### Testing

Unit tests should still be passing.

### Checklist

### DIFF
--- a/lib/psSupported.js
+++ b/lib/psSupported.js
@@ -1,3 +1,0 @@
-var semver = require('semver');
-
-module.exports = semver.satisfies(process.version, '^6.12.0 || >=8.0.0');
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -17,55 +17,43 @@ commands:
           command: npm test
 
 jobs:
```
RDIL, line 19: For the CI config, I've opted to change it to primarily using even-numbered releases, as they are the long-term-support versions, whereas the odd numbers are essentially development versions.  It doesn't really make sense to test against, say node 9 or 11, the majority of people use the stable versions such as 10 or 12.

PR #815, opened 2022-03-08
Support named exports in esm modules

perrin4869:
### Description

This PR adds the ability to use named exports in `type: module` node environments:

```js
import { sign } from "jsonwebtoken";
```

It accomplishes this via https://github.com/nodejs/cjs-module-lexer

### Testing

I added a test for this - unfortunately it will fail the lint parsing because it still supports node versions < 12.
Once the linting is updated, the tests will pass in all `node >= 12`.

### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,12 +1,6 @@
-module.exports = {
-  verify: require('./verify'),
-  sign: require('./sign'),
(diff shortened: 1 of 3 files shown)


perrin4869:
@jakelacey2012 any chance you can take a look at this?

PR #840, opened 2022-08-03
fix!: Prevent misconfiguration of the secret used with the HS*** algorithms

david-renaud-okta:
### Description

Added checks to make sure that a valid secret is used with the HS*** algorithms when signing and verifying

BREAKING CHANGE: Requires node 12.x or later to allow use of `KeyObject`

### Testing

- All automated tests pass
- Checked via script that this type of misconfiguration does not pass signing or verification

### Checklist


david-renaud-okta:
Closed in favour of https://github.com/auth0/node-jsonwebtoken/pull/852

Issue #711, opened 2020-04-20
RFC confrontation on "none" algo


frankli0324:
### Description

According to https://tools.ietf.org/html/rfc7518#section-3.6, jws s "MUST NOT accept such objects as valid unless the application specifies that it is acceptable for a specific object to not be integrity protected"

### Reproduction

```javascript
> var token = jwt.sign({'a':1}, 'asdf', { algorithm: 'none'})
undefined
> jwt.verify(token)
{ a: 1, iat: 1587359376 }
```

maybe we should add some kind of switch options?

### Expected output:

```javascript
> var token = jwt.sign({'a':1}, 'asdf', { algorithm: 'none'})
undefined
> jwt.verify(token)
throws error
> jwt.verify(token, '', { algorithm: 'none'})
{ a: 1, iat: 1587359376 }
```

frankli0324:
### Expected output:

```javascript
> var token = jwt.sign({'a':1}, 'asdf', { algorithm: 'none'})
undefined
> jwt.verify(token)
throws error
> jwt.verify(token, '', { algorithm: 'none'})
{ a: 1, iat: 1587359376 }
```

frankli0324:
https://github.com/auth0/node-jsonwebtoken/blob/5f10bf9957a2541828501cfecab0310908b2f62f/verify.js#L109-L111

so it's intentional?

panva:
@frankli0324 `none` won't be automatically accepted unless the developer provided no `keyOrSecret`. The three lines in isolation aren't the whole story.

https://github.com/auth0/node-jsonwebtoken/blob/5f10bf9957a2541828501cfecab0310908b2f62f/verify.js#L101-L118

frankli0324:
from my point of view, providing no secret doesn't mean that one explicitly "specifies that it is acceptable for a specific object to not be integrity protected", more like a mistake one could jump into

panva:
We'll consider making the `algorithms` option required when accepting unsigned JWTs for the next major.

frankli0324:
seems that this issue is forgotten?

PR #851, opened 2022-11-10
fix(sign&verify)!: Remove default `none` support from `sign` and `verify` methods, and require it to be explicitly configured

jakelacey2012:
### Description

This PR removes default support for the `none` algorithm for the `sign` and `verify` methods, now you have to specify `none` before too can `sign` and `verify`. This is to fix the issue raised in #711, but still allow users to use `none` if they wish to do so for development/testing.

BREAKING CHANGE: Removes fallback for `none` algorithm for the `verify` method.

### References

- #711 

### Testing

Please download this version of the package and then run the following code examples, which you can do by running `git@github.com:jakelacey2012/node-jsonwebtoken.git#IPS-2481`.

#### Simple test to get some background, should not throw... should not be effected by this change.
```javascript
var jwt = require('jsonwebtoken');
const decoded = jwt.verify(jwt.sign({ foo: 'bar' }, 'sshhhhhh'), 'sshhhhhh');
// should decode and not error.
```

#### Test sign method with algorithm `none` option,  should not error and return unsigned token.

```javascript
var jwt = require('jsonwebtoken');
const sign_none = jwt.sign({ foo: 'bar' }, undefined, { algorithm: 'none' });
// return unsigned token.
```

#### Test sign method with algorithm `none`, and not specify `none` in options.

```javascript
var jwt = require('jsonwebtoken');

jwt.sign({ foo: 'bar' }, undefined, { }); 
// this returns an error `secretOrPrivateKey must have a value`

jwt.sign({ foo: 'bar' }, 'secret', { }); 
// no error returned but token returned is signed with HS256

```

#### Testing verifying without `none` specified, should throw error `please specify "none" in "algorithms" to verify unsigned tokens`

```javascript
const unsigned = jwt.sign({ foo: 'bar' }, 'secret', { algorithm: 'none' });
const decoded = jwt.verify(unsigned, undefined, { algorithms: [] });
// should throw error 'please specify "none" in "algorithms" to verify unsigned tokens'
```

#### Testing verifying with `none` specified, should not throw and return a decoded token.

```javascript
const unsigned = jwt.sign({ foo: 'bar' }, 'secret', { algorithm: 'none' });
const decoded = jwt.verify(unsigned, undefined, { algorithms: ['none'] });
// should not throw and return a decoded token.
```

>
>

### Checklist

- [ ] The correct base branch is being used, if not the default branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,6 @@ encoded private key for RSA and ECDSA. In case of a private key with passphrase 
 `options`:
 
 * `algorithm` (default: `HS256`)
-  * * `none` MUST be configured in order to create unsigned tokens.
 * `expiresIn`: expressed in seconds or a string describing a time span [vercel/ms](https://github.com/vercel/ms). 
   > Eg: `60`, `"2 days"`, `"10h"`, `"7d"`. A numeric value is interpreted as a seconds count. If you use a string be sure you provide the time units (days, hours, etc), otherwise milliseconds unit is used by default (`"120"` is equal to `"120ms"`).
 * `notBefore`: expressed in seconds or a string describing a time span [vercel/ms](https://github.com/vercel/ms). 
@@ -139,7 +138,6 @@ As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues
 `options`
 
 * `algorithms`: List of strings with the names of the allowed algorithms. For instance, `["HS256", "HS384"]`.
-  * `none` MUST be configured in order to verify unsigned tokens.
 * `audience`: if you want to check audience (`aud`), provide a value here. The audience can be checked against a string, a regular expression or a list of strings and/or regular expressions. 
   > Eg: `"urn:foo"`, `/urn:f[o]{2}/`, `[/urn:f[o]{2}/, "urn:bar"]`
 * `complete`: return an object with the decoded `{ payload, header, signature }` instead of only the usual content of the payload.
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ encoded private key for RSA and ECDSA. In case of a private key with passphrase 
 `options`:
 
 * `algorithm` (default: `HS256`)
+  * * `none` MUST be configured in order to create unsigned tokens.
 * `expiresIn`: expressed in seconds or a string describing a time span [vercel/ms](https://github.com/vercel/ms). 
   > Eg: `60`, `"2 days"`, `"10h"`, `"7d"`. A numeric value is interpreted as a seconds count. If you use a string be sure you provide the time units (days, hours, etc), otherwise milliseconds unit is used by default (`"120"` is equal to `"120ms"`).
 * `notBefore`: expressed in seconds or a string describing a time span [vercel/ms](https://github.com/vercel/ms). 
@@ -138,6 +139,7 @@ As mentioned in [this comment](https://github.com/auth0/node-jsonwebtoken/issues
 `options`
 
 * `algorithms`: List of strings with the names of the allowed algorithms. For instance, `["HS256", "HS384"]`.
+  * `none` MUST be configured in order to verify unsigned tokens.
 * `audience`: if you want to check audience (`aud`), provide a value here. The audience can be checked against a string, a regular expression or a list of strings and/or regular expressions. 
   > Eg: `"urn:foo"`, `/urn:f[o]{2}/`, `[/urn:f[o]{2}/, "urn:bar"]`
 * `complete`: return an object with the decoded `{ payload, header, signature }` instead of only the usual content of the payload.
--- a/test/async_sign.tests.js
+++ b/test/async_sign.tests.js
@@ -41,16 +41,6 @@ describe('signing a token asynchronously', function() {
       });
     });
 
-    //Known bug: https://github.com/brianloveswords/node-jws/issues/62
-    //If you need this use case, you need to go for the non-callback-ish code style.
-    it.skip('should work with none algorithm where secret is falsy', function(done) {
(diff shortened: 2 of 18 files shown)


david-renaud-okta:
Is this test still necessary? I think we should enable (with appropriate modifications) or remove this test as part of this change.

https://github.com/jakelacey2012/node-jsonwebtoken/blob/IPS-2481/test/async_sign.tests.js#L44-L52

david-renaud-okta (changes requested):

test/async_sign.tests.js
```diff
@@ -34,7 +34,7 @@ describe('signing a token asynchronously', function() {
     });
 
     it('should work with none algorithm where secret is set', function(done) {
```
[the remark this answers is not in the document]
jakelacey2012: Ahh good catch actually I think we can remove this test because we already test this [as part of the schema here](https://github.com/auth0/node-jsonwebtoken/blob/d3b8c1d1538800f5fa9536a6419e0ec8761d193b/test/schema.tests.js#L22-L24).
https://github.com/auth0/node-jsonwebtoken/blob/d3b8c1d1538800f5fa9536a6419e0ec8761d193b/test/schema.tests.js#L22-L24

david-renaud-okta (approved):

test/verify.tests.js
```diff
@@ -30,39 +30,52 @@ describe('verify', function() {
...
       header: header,
       payload: payload,
-      secret: priv,
+      secret: 'secret',
       encoding: 'utf8'
     });
 
-    jwt.verify(signed, null, {typ: 'JWT'}, function(err, p) {
+    expect(function () {
+      jwt.verify(signed, 'secret', {typ: 'JWT'});
```
CharlesRea, line 45: This case would already return an error before your changes, as you're passing in a secret and we do this check:
```
    if (!hasSignature && secretOrPublicKey){
    }
```
We should have a test for both cases where secret is provided, or is null.

verify.js
```diff
@@ -106,10 +112,6 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
       return done(new JsonWebTokenError('secret or public key must be provided'));
     }
 
-    if (!hasSignature && !options.algorithms) {
-      options.algorithms = ['none'];
-    }
```
CharlesRea, line 111: If `secretOrPublicKey` and `options.algorithms` are both null / undefined, line 116 will error. This was previously not possible as if `secretOrPublicKey` is undefined then `hasSignature` is guaranteed to be false (per check inline 111), which means `options.algorithms` was always set to `[none]`.

If `alg === none` then we'll have already failed, but if `alg !== none` then we'll get an `can't read toString() of undefined` error. It is correct that the token isn't valid in that case, but we should return an actual invalid token error (from passing this through to `jws.verify`).

IMO, we should also avoid changing the error message unnecessarily if we were previously going to be caught by the two checks above.

So what do you think about removing the check you've added on line 83, and putting the check here instead with the old condition. This will still prevent `none` algs being verified without being specified, as we no longer default to `none` in this case:

```js
if (!hasSignature && !options.algorithms) {
    return done(new JsonWebTokenError('specify "none" in "algorithms" to verify unsigned tokens'));
}
```

test/async_sign.tests.js
```diff
@@ -41,16 +41,6 @@ describe('signing a token asynchronously', function() {
       });
     });
 
-    //Known bug: https://github.com/brianloveswords/node-jws/issues/62
```
CharlesRea, line 44: Did we mean to remove this test?

test/claim-aud.test.js
```diff
@@ -6,7 +6,7 @@ const util = require('util');
 const testUtils = require('./test-utils');
 
 function signWithAudience(audience, payload, callback) {
-  const options = {algorithm: 'none'};
+  const options = {algorithm: 'HS256'};
```
CharlesRea, line 9: We may be able to revert all these test changes with the latest change to `sign`. What do you think?

verify.js
jakelacey2012: Good catch, yeah what you've said makes sense let's move the check there.

test/async_sign.tests.js
jakelacey2012: This is something I changed as part of [David's comment](https://github.com/auth0/node-jsonwebtoken/pull/851#issuecomment-1314319951) and was changed when we were removing the `none` functionality all together, since we're not doing that anymore I think we can just add this back in.

test/claim-aud.test.js
jakelacey2012: We would be able to revert but the places that we use none and verify we would need to configure that algorithm explicitly, do you see an advantage in using `none` in these tests or was comment more about making the PR as small as possible?

I think I prefer using `HS256` for example rather than `none`, since its something we don't want to encourage and should be used in exceptional cases. Also means in the tests we just configure the algorithm, the secret and no options for the verify - the alternative seems verbose and repetitive IMO.

Let me know what you think :)

README.md
```diff
@@ -38,6 +38,7 @@ encoded private key for RSA and ECDSA. In case of a private key with passphrase
 `options`:
 
 * `algorithm` (default: `HS256`)
+  * * `none` MUST be configured in order to create unsigned tokens.
```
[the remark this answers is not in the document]
CharlesRea: I'm not sure we want to make it easy to create `none` tokens :P I'd remove this personally, I think it adds confusion to someone wanting to use the library who isn't hugely experienced with best practices here - let's not call out `none` specifically.

test/claim-aud.test.js
CharlesRea: Yep agreed :)

CharlesRea (approved):

PR #852, opened 2022-11-16
fix!: Prevent accidental use of insecure key sizes & misconfiguration of secrets

david-renaud-okta:
- Added checks to prevent invalid secrets from being used with the HS*** algorithms when signing and verifying
- Added checks to prevent the use of insecure asymmetric key sizes except when explicitly overriden via options

BREAKING CHANGE: Requires node 12.x or later to allow use of `KeyObject`

### Description

This PR covers addresses 3 misconfigurations which could compromise security.
1. Signing with a public key that has a modulus length of less than 2048 is prevented by default. If required this check can be disabled by setting `allowInsecureKeySizes` to `true` in the options object.
2. It is no longer possible to use an RSA key to verify an HMAC signature
3. Prevented `Buffer`s containing malicious objects from being used as key material.

### Testing

New automated tests have been added for each scenario.

### Checklist

### DIFF
--- a/test/jwt.malicious.tests.js
+++ b/test/jwt.malicious.tests.js
@@ -1,0 +1,32 @@
+const jwt = require('../index');
+const crypto = require("crypto");
+const {expect} = require('chai');
+const JsonWebTokenError = require("../lib/JsonWebTokenError");
+
+describe('when verifying a malicious token', function () {
+  // attacker has access to the public rsa key, but crafts the token as HS256
+  // with kid set to the id of the rsa key, instead of the id of the hmac secret.
+  // const maliciousToken = jwt.sign(
+  //   {foo: 'bar'},
+  //   pubRsaKey,
+  //   {algorithm: 'HS256', keyid: 'rsaKeyId'}
+  // );
+  // consumer accepts self signed tokens (HS256) and third party tokens (RS256)
+  const options = {algorithms: ['RS256', 'HS256']};
+
+  const {
+    publicKey: pubRsaKey
+  } = crypto.generateKeyPairSync('rsa', {modulusLength: 2048});
+
+  it('should not allow HMAC verification with an RSA key in KeyObject format', function () {
+    const maliciousToken = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6InJzYUtleUlkIn0.eyJmb28iOiJiYXIiLCJpYXQiOjE2NTk1MTA2MDh9.cOcHI1TXPbxTMlyVTfjArSWskrmezbrG8iR7uJHwtrQ';
+
+    expect(() => jwt.verify(maliciousToken, pubRsaKey, options)).to.throw(JsonWebTokenError, 'must be a symmetric key');
+  })
+
+  it('should not allow HMAC verification with an RSA key in PEM format', function () {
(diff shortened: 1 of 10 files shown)


jakelacey2012 (changes requested):
Looks like we have missing coverage for node-16 and 18.
```
ERROR: Coverage for branches (94.72%) does not meet global threshold (95%)
```

However the changes in general look fine to me 👍 think we just need to find where that missing coverage is and add the tests.

jakelacey2012 (approved):

verify.js
```diff
@@ -110,18 +111,39 @@ module.exports = function (jwtString, secretOrPublicKey, options, callback) {
...
+    }
 
+    if (!options.algorithms) {
+      if (secretOrPublicKey.type === 'secret') {
+        options.algorithms = HS_ALGS;
+      } else if (secretOrPublicKey.asymmetricKeyType === 'rsa') {
+        options.algorithms = RSA_KEY_ALGS
+      } else {
+        options.algorithms = PUB_KEY_ALGS
+      }
```
CharlesRea, line 133: This is a slight change from previous behaviour: if we had a PEM with `BEGIN RSA PUBLIC KEY`, we knew we could narrow down to RSA only, otherwise for `BEGIN PUBLIC KEY` we allow all public algs.

The former case could have `asymmetricKeyType === 'pub-rsa'` so we end up allowing slightly more than before. 

IMO let's pick up RSA-PSS here, and also handle ECC while we're here for consistency (technically a breaking change for EC though so we need to document it):

```
      if (secretOrPublicKey.type === 'secret') {
        options.algorithms = HS_ALGS;
      } else if (secretOrPublicKey.asymmetricKeyType === 'rsa-pss' || secretOrPublicKey.asymmetricKeyType === 'rsa-pss') {
        options.algorithms = RSA_KEY_ALGS;
      } if (secretOrPublicKey.asymmetricKeyType === 'ec') {
        options.algorithms = EC_KEY_ALGS;
      }
       else {
        options.algorithms = PUB_KEY_ALGS;
      }
```

CharlesRea: It'd be great to add some explanation of this in the README as well - we don't specify what happens if `algorithms` is undefined

test/jwt.malicious.tests.js
```diff
@@ -0,0 +1,32 @@
...
+  // );
+  // consumer accepts self signed tokens (HS256) and third party tokens (RS256)
+  const options = {algorithms: ['RS256', 'HS256']};
+
+  const {
+    publicKey: pubRsaKey
+  } = crypto.generateKeyPairSync('rsa', {modulusLength: 2048});
+
+  it('should not allow HMAC verification with an RSA key in KeyObject format', function () {
+    const maliciousToken = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6InJzYUtleUlkIn0.eyJmb28iOiJiYXIiLCJpYXQiOjE2NTk1MTA2MDh9.cOcHI1TXPbxTMlyVTfjArSWskrmezbrG8iR7uJHwtrQ';
```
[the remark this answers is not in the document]
david-renaud-okta: This token does not have an expiry. `iat` only affects the outcome if `maxAge` is set in `options` which is not used for this test. We are good here.

verify.js
david-renaud-okta: Done. We'll add the change to our migration guide

PR #879, opened 2023-01-12
remove lodash

### LINKED ISSUE #878: Version 9.0.0 Increased bundle size, probably due to lodash import
### Description

When trying to update to 9.0.0 we noticed that the bundle size of jsonwebtoken had increased quite a lot, making a small app of ours exceed a max size limit at a cloud platform.

It seems to us at least that it might be due to importing the whole lodash package in package.json instead of just the functions used, introduced in this commit: ecdf6cc6073ea13a7e71df5fad043550f08d0fa6

I realise this doesn't necessarily fit all that well under "bug", sorry in that case.

### Reproduction
Here's two screenshots from bundlephobia comparing 8.5.1 vs 9.0.0

**8.5.1:**
image
image

**9.0.0:**
image
image

### Environment
- **Version of this library used:** 9.0.0

Uzlopak:
Resolves #878 

### Description

>
>
> If the UI is being changed, please provide screenshots.

### References

>
> - Auth0 Community post
> - StackOverflow post
> - Support forum thread
> - Related pull requests/issues from other repos
>
> If there are no references, simply delete this section.

### Testing

>
>

### Checklist

- [ ] The correct base branch is being used, if not the default branch


KuSh:
Reverting to per module packages as it was before https://github.com/auth0/node-jsonwebtoken/commit/ecdf6cc6073ea13a7e71df5fad043550f08d0fa6 could do the  same without reimplementing the wheel

Uzlopak:
I think it is pure garbage to use lodash at all. And the only reason I "reinvented" the wheel was because lodash garbage was used in the first place and there could be some maniac using Number-classes and it would be backwards breaking to not use that isNumber functionality. The same with once and isPlainObject. It is only for backwards compatibility.

KuSh:
> I think it is pure garbage to use lodash at all. And the only reason I "reinvented" the wheel was because lodash garbage was used in the first place and there could be some maniac using Number-classes and it would be backwards breaking to not use that isNumber functionality. The same with once and isPlainObject. It is only for backwards compatibility.

I'm not found of lodash neither but it as the avantage of being tested and used by many, contrary to your implementations. As the goal is to reduce final package sise importing individual methods will accomplish the same without maintenance burden nor testing on node-jsonwebtoken side but with the guarantee that those methods are well tested

araphiel-chwy (approved):

araphiel-chwy (changes requested):
My apologies for the accidental approval.

There might be some hesitancy with this approach - so here's a lower impact alternative.

Uzlopak:
Closing due to inactivity

PR #889, opened 2023-02-22
Reduce import cost of lodash

araphiel-chwy:
### Description

Use direct lodash imports to reduce bundle sizes.

### Checklist


jimmywarting:
I think most of this can be inlined (coded by hand / plain vanilla javascript)

araphiel-chwy:
@jimmywarting Yep, I agree. 

But this approach being is used to reduce any potential impact of these change for the maintainers.

Related: #887 #879

skvale:
I just wanted to chime in that this is important to me

node-jsonwebtoken is the only issuer path that we have that requires `lodash` without the slash
Screenshot 2023-03-10 at 11 30 30 AM


And here's the impact on our bundle size, 95kb (gzip)
Screenshot 2023-03-10 at 11 30 49 AM

araphiel-chwy:
Just wanted to flag this for review @CharlesRea @jakelacey2012 

This should address bundle size issues for people using parts of the `node-jsonwebtoken` in their webpack builds

PR #946, opened 2023-11-03
fix: `validator.isValid is not a function` for certain objects

### LINKED ISSUE #945: `sign` crashes with `validator.isValid is not a function` for certain objects
### Description

> Provide a clear and concise description of the issue, including what you expected to happen.

I noticed that every once in a blue moon, my tests for something that calls `sign` would fail with `validator.isValid` is not a function. After debugging, I managed to narrow it down. I'm using [fast-check](https://fast-check.dev) for tests, and their `fc.object` function sometimes generates objects that have keys like `"__proto__"`, `"valueOf"`, or `"toString"`. Whenever an object like this is passed into `sign`, the `validator.isValid` error appears.

### Reproduction

Most minimal reproduction I could create:

1. Create a new Node project, installing `jsonwebtoken`
2. In a file `index.js`, add:

```javascript
const { sign } = require("jsonwebtoken");

console.log(sign({ valueOf: 0 }, "anysecret");
```

3. Run `node index.js`

To reproduce the `fast-check` test that shows the different ways this error occurs: 

1. Create a new Node project, installing `jest`, `jsonwebtoken`, and `fast-check`.
2. In a file `jwt.test.js`, add:

```javascript
const fc = require("fast-check");
const { sign } = require("jsonwebtoken");

describe("jwt.sign", () => {
  it("should sign any object", () => {
    const prop = fc.property(
      fc.object(),
      fc.base64String({ minLength: 1 }),
      (obj, secret) => {
        const result = sign(obj, secret);
        return !!result;
      },
    );

    fc.assert(prop, { numRuns: 1000, verbose: true });
  });
});
```

3. Run `npx jest`
4. See that the test does not pass for certain values, and a `TypeError: validator.isValid is not a function` is thrown

### Environment

- **Version of this library used:** `^9.0.2`
- **Version of the platform or framework used, if applicable:** N/A
- **Other relevant versions (language, server software, OS, browser):** both JS and TS, on Linux, using Node
- **Other modules/plugins/libraries that might be involved:** `fast-check`

ej-shafran:
### Description

Objects with keys like `valueOf`, `toString`, and `__proto__` cause a `TypeError` to be raised when calling `jwt.sign`. This is because the key technically does exist on the `schema` param of `validate`, when checked with `if (schema[key]) {}`. Using
`Object.prototype.hasOwnProperty` solves the issue.

I used `Object.prototype.hasOwnProperty.call(schema, key)`. This method has full browser support. Alternatively, `lodash.has` could be installed and used, but I think that would be overkill.

### References

>
> GitHub Issue/PR number addressed or fixed

Fixes #945

### Testing

I have added `test/issue_945.tests.js`. It fails if the change made to `sign.js` is not present.

### Checklist

### DIFF
--- a/test/issue_945.tests.js
+++ b/test/issue_945.tests.js
@@ -1,0 +1,12 @@
+const jwt = require("..");
+
+const KEY = "any_key";
+
+describe("issue 945 - validator.isValid is not a function", () => {
+  it("should work", () => {
+    jwt.sign({ hasOwnProperty: null }, KEY);
+    jwt.sign({ valueOf: null }, KEY);
+    jwt.sign({ toString: null }, KEY);
+    jwt.sign({ __proto__: null }, KEY);
+  });
+});
--- a/sign.js
+++ b/sign.js
@@ -45,13 +45,13 @@ function validate(schema, allowUnknown, object, parameterName) {
   }
   Object.keys(object)
     .forEach(function(key) {
-      const validator = schema[key];
-      if (!validator) {
+      if (!Object.prototype.hasOwnProperty.call(schema, key)) {
         if (!allowUnknown) {
           throw new Error('"' + key + '" is not allowed in "' + parameterName + '"');
         }
         return;
       }
+      const validator = schema[key];
       if (!validator.isValid(object[key])) {
         throw new Error(validator.message);
       }


test/issue_945.tests.js
```diff
@@ -0,0 +1,12 @@
+const jwt = require("..");
+
+const KEY = "any_key";
+
+describe("issue 945 - validator.isValid is not a function", () => {
```
dubzzz, line 5: As the issue has been detected via property based testing mechanisms, I wonder if at some point having such tests to cover this project could be a thing. It seems that it could be useful in addition to these tests to avoid regressions or even more nasty bugs such as things related to nested objects...

Issue #952, opened 2023-12-22
Rename `decode` function `unsafeDecode` to highlight the risk


marine-mb:
### Describe the problem you'd like to have solved

I'm a DevSec. I help developers teams find and fix their vulnerabilities.
In two of my last missions, I've seen developers using `decode` instead of `verify` even if the documentation has been improved. 

### Describe the ideal solution

To avoid this and reduce the vulnerable apps, I think it would be great to rename the `decode` function and call it `unsafeDecode`. (as suggested by @youssef-jbili)

Other libraries have done it and it helps developers reduce their mistakes:
- `dangerouslySetInnerHTML` for React
- `bypassSecurityTrustHtml` for Angular

I can make a Pull Request if you think it is a good idea.

youssef-jbili:
Very interesting proposal. I would just suggest calling it `unsafeDecode` instead to maintain the same camel case naming

jonaskello:
I would suggest `unverifiedDecode` to keep it within the same wording as `verify()` function. I think it also would be nice to be able to take this unverified decoded token and verify it (see #955). Code would then be something like:

```js
const unverifiedToken = unverifiedDecode(rawToken);
const verifiedToken = verifyDecoded(unverifiedToken);
```

The current `verify()` function actually does decoding also so a better name for that IMO would be `verifiedDecode()`. So IMO the ideal API would be:

- `unverifiedDecode(rawToken)` - this is the same as current `decode()`
- `verifiedDecode(rawToken, ...)`  - this is the same as current `verify()`
- `verifyDecoded(unverifiedDecodedToken, ...)` - new function to just verify an unverified but decoded token (returns Promise)

marine-mb:
Hi @jonaskello,

I think `unverifiedDecode` is less clear for developers. 
It should be evident that it is dangerous! 

Also, verifying a token after it has been decoded is impossible. 
The JWT is split into three parts: header, payload and signature. If the JWT is modified, the signature is not valid anymore.
When you decode the JWT, you only get the payload (without the base64 encoding) so you can't verify the signature anymore.
I hope it's clear.

@aalu-love, thanks for the PR. Would you update it to use camelCase as @youssef-jbili suggested? Otherwise, I can do it at the end of the week 😁

jonaskello:
@marine-mb When you decode you get all three parts as can be seen [here](https://github.com/auth0/node-jsonwebtoken/blob/master/decode.js#L23)

```ts
return {
[...]

PR #953, opened 2023-12-30
Issue #952 fixed rename decode function unsafe_decode

aalu-love:
#952 I have made the changes as per the above mentioned in the above comment.

### Description

> Renamed decode function to unsafe_decode where ever it has define in the code.

### Checklist


README.md
```diff
@@ -243,7 +243,7 @@ jwt.verify(token, getKey, options, function(err, decoded) {
 <details>
 <summary><em></em>Need to peek into a JWT without verifying it? (Click to expand)</summary>
 
-### jwt.decode(token [, options])
+### jwt.unsafe_decode(token [, options])
```
marine-mb, line 246: @aalu-love Could you make it camelCase? So it can match javascript default case

> 

> 

>

package-lock.json
```diff
@@ -0,0 +1,7268 @@
+{
+  "name": "jsonwebtoken",
```
marine-mb, line 2: It seems the package-lock wasn't committed before which is not the best practice. 
It might be better to add it in a different commit for visibility

PR #972, opened 2024-06-09
Optionally validate the payload of a token before verification

### LINKED ISSUE #955: Verify an already decoded token
### Describe the problem you'd like to have solved

Be able to verify an already decoded token.

We sometimes decode the token before verification to check if certain claims are present. For example tenant claim of the token is used to determine which jwksuri to use for verification. Also the kid is in the header of the decoded token and is needed to fetch correct key for verification.

Today the verify() function first call decode() internally and then continues to do the actual verification. I would be nice if verify() only did the actual verification and let the caller do the decode().

### Describe the ideal solution

verify() function would take a decoded token as argument, or a new function verifyDecoded() could be introduced if overloading is not wanted.

## Alternatives and current work-arounds

We could decode the token to get the pre-verify claims and then call verify which then again decodes the token. This works but puts overhead on every request as we always verify the token.

georgejmx:
### Description

>

Currently the only way to know that a token returned from `jwt.verify()` is of the correct format is by verifying the payload returned from the function. This MR makes it so the user can specify a callback `payloadCallback` that can be provided as an argument to `jwt.verify()`, which facilitates sanitising the token payload before it is verified

### References

>

Closes #955 

### Testing

>
>

- [x] This change adds full test coverage for new/changed/fixed functionality
- [x] This change was made on Ubuntu with Node 20.14

### Checklist

- [x] I have added documentation for new/changed functionality in the README

### DIFF
--- a/test/payload_callback.test.js
+++ b/test/payload_callback.test.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const jwt = require('../index');
+const assert = require('chai').assert;
+const expect = require('chai').expect;
+
+describe('payload callback', function() {
+  const KEY = 'somethingSECRET';
+  const TEST_ERROR_MESSAGE = 'bar not car!';
+  let testPayloadCallback;
+
+  beforeEach(function() {
+    testPayloadCallback = function (payload) {
+      if (payload.foo !== 'bar') {
+        throw new Error(TEST_ERROR_MESSAGE);
+      }
+    }
+  });
+
+  it('should check that the payload satisfies the provided callback', function () {
+    const token = jwt.sign({ foo: 'bar' }, KEY);
+    const result = jwt.verify(token, KEY, undefined, undefined, testPayloadCallback);
+    expect(result.foo).to.equal('bar');
+  });
+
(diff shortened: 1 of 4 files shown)


test/payload_callback.test.js
```diff
@@ -0,0 +1,44 @@
...
+    testPayloadCallback = function (payload) {
+      if (payload.foo !== 'bar') {
+        throw new Error(TEST_ERROR_MESSAGE);
+      }
+    }
+  });
+
+  it('should check that the payload satisfies the provided callback', function () {
+    const token = jwt.sign({ foo: 'bar' }, KEY);
+    const result = jwt.verify(token, KEY, undefined, undefined, testPayloadCallback);
```
georgejmx, line 22: Not keen on the double `undefined`, however had concerns around making any other changes to the function signature wrt backward compatibility

georgejmx: Here is a repo that demos the feature https://github.com/georgejmx/demo-jsonwebtoken-payload-callback

Any more manual testing you need just let me know @Vero7979

Issue #875, opened 2023-01-06
v9: `decode` does not work with `import *` in TypeScript


bluepichu:
### Description

If you `import * as jwt from "jsonwebtoken"`, then you can't call `jwt.decode`.  It works fine if you use a default import though (`import jwt from "jsonwebtoken"`).

Related to #866, and probably also #870.

The root cause of this is that `import *` in TS gets translated in such a way that only enumerable properties are copied from the module.  Since `decode` is non-enumerable it is not properly copied.

### Reproduction

```ts
import * as jwt from "jsonwebtoken";

jwt.decode("abc"); // TypeError: jwt.decode is not a function
```

### Environment

Version 9.0.0 on Node 16.19.0 with TS 4.6.2

swapnil-sudhir:
Any solution for this so far ? @bluepichu

lloydjatkinson:
```ts
import { decode } from 'jsonwebtoken';

console.log(
    decode(
        'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJPbmxpbmUgSldUIEJ1aWxkZXIiLCJpYXQiOjE2NzMyNzcwNTMsImV4cCI6MTcwNDgxMzA1MywiYXVkIjoid3d3LmV4YW1wbGUuY29tIiwic3ViIjoianJvY2tldEBleGFtcGxlLmNvbSIsIkdpdmVuTmFtZSI6IkpvaG5ueSIsIlN1cm5hbWUiOiJSb2NrZXQiLCJFbWFpbCI6Impyb2NrZXRAZXhhbXBsZS5jb20iLCJSb2xlIjpbIk1hbmFnZXIiLCJQcm9qZWN0IEFkbWluaXN0cmF0b3IiXX0.cN6HdDt31tosRDffE6UIgDGy3lgjFW9j9uO-VCd8S7U'
    )
);
```

rpuls:
I am having an issue that I think is related. Long story short, I am getting: `TypeError: Cannot redefine property: decode` when trying to stub `decode` with sinon.stub - but no issue stubbing the `verify` function.

I posted the issue on OS with some more detail:
https://stackoverflow.com/questions/75071217/jsonwebtoken-9-0-0-getting-error-typeerror-cannot-redefine-property-decode

PR #997, opened 2025-08-08
Update export styles for ESM compatibility

### LINKED ISSUE #963: import `jsonwebtoken` throws error
### Description

import `jsonwebtoken` throws error

### Reproduction

The code:
```javascript
import { verify } from "jsonwebtoken";
```

The error message:
```console
import { verify } from "jsonwebtoken";
         ^^^^^^
SyntaxError: Named export 'verify' not found. The requested module 'jsonwebtoken' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'jsonwebtoken';
const { verify } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:134:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:217:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:120:12)

Node.js v21.7.0
```

It works on v9.0.0 but fails on 9.0.2.
### Environment

- **Version of this library used:** 9.0.2
- **Version of the platform or framework used, if applicable:**
- **Other relevant versions (language, server software, OS, browser):**
- **Other modules/plugins/libraries that might be involved:**

joshkel:
### Description

Node.js's ESM-to-CJS interoperability only supports a limited subset of exports.  (See [cjs-module-lexer](https://github.com/nodejs/cjs-module-lexer/blob/main/README.md) for details.)  By switching to one of the supported styles, ESM code can use named imports from jsonwebtoken.

### References

This issue is further documented as the ["Named Exports"](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/NamedExports.md) error on [Are the Types Wrong](https://arethetypeswrong.github.io/?p=jsonwebtoken%409.0.2).

Fixes #963, #875

### Testing

To demonstrate, create a `test.mjs`:

```js
// Within the jsonwebtoken directory:
import { verify } from './index.js';
// Or importing from an installed jsonwebtoken package:
import { verify } from 'jsonwebtoken';
verify('a', 'a');
```

Without this change:

```
import { verify } from './index.js';
         ^^^^^^
SyntaxError: Named export 'verify' not found. The requested module './index.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:
```

With this change: Successful execution (in this case, a "jwt malformed" error).

Tested in Node.js 24.4.1

### Checklist

- [X] The correct base branch is being used, if not the default branch

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,15 @@
+const decode = require('./decode');
+const verify = require('./verify');
+const sign = require('./sign');
+const JsonWebTokenError = require('./lib/JsonWebTokenError');
+const NotBeforeError = require('./lib/NotBeforeError');
+const TokenExpiredError = require('./lib/TokenExpiredError');
+
 module.exports = {
-  decode: require('./decode'),
-  verify: require('./verify'),
-  sign: require('./sign'),
-  JsonWebTokenError: require('./lib/JsonWebTokenError'),
-  NotBeforeError: require('./lib/NotBeforeError'),
-  TokenExpiredError: require('./lib/TokenExpiredError'),
+  decode,
+  verify,
+  sign,
+  JsonWebTokenError,
+  NotBeforeError,
+  TokenExpiredError,
 };


squirly:
I tested this patch in my codebase and it fixes imports for this library when using ESM.

Issue #1000, opened 2025-09-10
`jwt.sign()` callback is executed twice for "The payload already has an "..." property" errors


pfgithub:
### Description

The callback to `jwt.sign()` is executed twice when there is an error `Bad "options.issuer" option. The payload already has an "iss" property`. I would expect the callback to only be executed once.

### Reproduction

```js
// a.mjs
import jwt from "jsonwebtoken";
jwt.sign(
  {
    iss: "bar",
    iat: 1757476476,
  },
  "secret",
  {
    algorithm: "HS256",
    issuer: "foo",
  },
  (err, asyncSigned) => {
    console.log("callback called:", asyncSigned ?? err?.message);
  },
);
```

```
$> npm init -y && npm add jsonwebtoken
$> node a.mjs
callback called: Bad "options.issuer" option. The payload already has an "iss" property.
callback called: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJiYXIiLCJpYXQiOjE3NTc0NzY0NzZ9.TkSecbZDocHnjwnuDhQRXNvRkxsERNGBkRc6P0wHlNY
```

Notice that "callback called" is printed twice, while I would expect it to be printed only once.

### Environment

jsonwebtoken 9.0.2 on Darwin 24.6.0 arm64 arm

### Other information

It looks like the error is here:

https://github.com/auth0/node-jsonwebtoken/blob/bc28861f1fa981ed9c009e29c044a19760a0b128/sign.js#L217-L225

Typically, `return failure()` throws which exits the forEach and the function. But when a callback is provided, it only returns from the forEach. The rest of the keys are checked and the rest of the function still runs.

maishivamhoo123:
@pfgithub I think you are right because forEach will not return even after getting an error.
I think we should use a for loop here.
Something like:
for (const key of Object.keys(options_to_payload)) {
//same logic 
}

This way the function will properly return after the first error and won’t call the callback twice.
Am I right?
@team can you please assign this issue to me

azender1:
Bugs like this are easy to underestimate because “callback executed twice” sounds cosmetic until the callback is wired to a mutation.

At that point it stops being just a library bug and becomes a duplicate-execution risk.

The safest pattern I’ve found around this class of problem is:

treat the callback-triggered mutation as a logical operation with a stable request ID
persist a durable execution receipt before / at the side-effect boundary
make re-entry return the prior result instead of executing again

[...]

PR #1002, opened 2025-11-07
fix(sign): prevent double callback when payload already has a property

maishivamhoo123:

### Description

This PR fixes a bug in jwt.sign() where the callback is called twice when the payload already contains a property that conflicts with options. For example, when payload.iss exists and options.issuer is also provided, the callback was triggered once with an error and then again with a token.

The root cause was that Object.keys(options_to_payload).forEach(...) with return failure(err) only returned from the callback, not the parent function, allowing the signing logic to continue and call the callback a second time.

Solution:

Replaced .forEach() with a for...of loop so return failure(err) exits the sign() function immediately.

Added a regression test to ensure the callback is called only once in conflict scenarios.

No breaking changes to the API.

References

Fixes GitHub issue: #1000 

Screenshot 2025-11-07 192917

>
> - Auth0 Community post
> - StackOverflow post
> - Support forum thread
> - Related pull requests/issues from other repos
>
> If there are no references, simply delete this section.

Added a regression test in test/callback-issue.test.js:

const assert = require('assert');
const jwt = require('../index.js');

describe('jwt.sign callback fix', function () {
  it('should call callback only once on payload/options conflict', function (done) {
    let callbackCount = 0;

    jwt.sign(
      { iss: 'bar', iat: 1757476476 },
      'secret',
      { algorithm: 'HS256', issuer: 'foo' },
      (err) => {
        callbackCount++;
        assert.ok(err);
        assert.strictEqual(
          err.message,
          'Bad "options.issuer" option. The payload already has an "iss" property.'
        );
        assert.strictEqual(callbackCount, 1);
        done();
      }
    );
  });
});

All existing tests pass.

Regression test verifies callback is called only once.

Developed and tested in Node.js v22.19.0 on Linux (WSL/Debian).

 This change adds test coverage for the fixed functionality.

Checklist

 I have added tests for the fixed functionality

 All active GitHub checks for tests, formatting, and security are passing

 The correct base branch is being used (master).

### DIFF
--- a/sign.js
+++ b/sign.js
@@ -214,7 +214,7 @@ module.exports = function (payload, secretOrPrivateKey, options, callback) {
     }
   }
 
-  Object.keys(options_to_payload).forEach(function (key) {
+  for (const key of Object.keys(options_to_payload)) {
     const claim = options_to_payload[key];
     if (typeof options[key] !== 'undefined') {
       if (typeof payload[claim] !== 'undefined') {
@@ -222,7 +222,7 @@ module.exports = function (payload, secretOrPrivateKey, options, callback) {
       }
       payload[claim] = options[key];
     }
-  });
+  };
 
   const encoding = options.encoding || 'utf8';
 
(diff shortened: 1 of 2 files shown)


maishivamhoo123:
@jpadilla @siacomuzzi and @team can you please review.

PR #1013, opened 2026-03-16
Reject non-finite exp/nbf/iat values and tighten validation

AzmeerX:
### Description

1. Reject non-finite registered claim values (exp, nbf, iat) by tightening validation to require finite numbers.
2. Align tests to assert errors for Infinity, -Infinity, and NaN values instead of silently coercing them.
3. Keeps current API intact while preventing tokens with invalid temporal claims from being produced.

### Testing

Commands run locally (Node v20.20.0 and above):
npm test -- test/claim-iat.test.js
npm test -- test/claim-nbf.test.js
npm test -- test/claim-exp.test.js
npm test

### Checklist

 All active GitHub checks for tests, formatting, and security are passing
 The correct base branch is being used, if not the default branch

### DIFF
--- a/sign.js
+++ b/sign.js
@@ -5,7 +5,6 @@ const jws = require('jws');
 const includes = require('lodash.includes');
 const isBoolean = require('lodash.isboolean');
 const isInteger = require('lodash.isinteger');
-const isNumber = require('lodash.isnumber');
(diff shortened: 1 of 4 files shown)


AzmeerX:
let me know if you have any query about the commit or need any further change

